PR #44, opened 2017-06-26
Add query expand command.

komainu8:


sql/function/query-expand/one.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE full_text_sarch_engine (
+  name text
+);
+
+INSERT INTO full_text_sarch_engine VALUES ('Groonga');
+INSERT INTO full_text_sarch_engine VALUES ('Senna');
+INSERT INTO full_text_sarch_engine VALUES ('Namazu');
+INSERT INTO full_text_sarch_engine VALUES ('Hyper Estraier.');
+
+SELECT pgroonga.query_expand('Groonga');
```
kou: We doesn't need `CREATE TABLE` and `INSERT`.

expected/function/query-expand/one.out
```diff
@@ -0,0 +1,6 @@
+SELECT pgroonga.query_expand('Groonga');
+ full_text_sarch_engine
+------------------------
+ Groonga
+ Senna
+ (2 row)
```
kou: Did PGroonga really output it? I don't know why multiple records are returned.

data/pgroonga.sql
```diff
@@ -2,6 +2,13 @@ SET search_path = public;
 
 CREATE SCHEMA pgroonga;
 
+CREATE FUNCTION pgroonga.query_expand(term text)
+	RETURNS text
+	AS 'MODULE_PATHNAME', 'pgroonga_query_expand'
+	LANGUAGE C
+	VOLATILE
```
kou: It should be `STABLE` for optimization.

komainu8:
Thank you for your comments.
I've fixed them.

kou:
I've merged.
Normally, "Fix review comments" isn't a good commit message. Commit message should describe summary of the changes or why the changes are valid.

See also (in Japanese):

  * http://www.clear-code.com/blog/2012/2/21.html
  * http://www.clear-code.com/blog/2013/4/24.html

PR #125, opened 2020-03-09
jsonb: fix a wrong how to use jsonb index

komainu8:
TODO:implementation

PGroonga occurs error when executing `SELECT COUNT(*) FROM xxx` if we set JSONB index and nothing `WHERE` phrase.

```sql
CREATE TABLE logs (
  record jsonb
);

CREATE INDEX pgroonga_index ON logs USING pgroonga (record);

SET enable_seqscan = false;
SELECT COUNT(*) FROM logs;

ERROR:  pgroonga: object isn't found: <Lexiconxxxxx_0.index>
```

### DIFF
--- a/expected/jsonb/count/count.out
+++ b/expected/jsonb/count/count.out
@@ -1,0 +1,11 @@
+CREATE TABLE logs (
+  record jsonb
+);
+CREATE INDEX pgroonga_index ON logs USING pgroonga (record);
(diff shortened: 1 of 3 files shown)


komainu8:
This issue already solved by 08da7e28abf96b4294031e75793cf51d34ac77b7.

PR #131, opened 2020-05-06
package fedora: add a package for Fedora

komainu8:


kou:
Added: https://github.com/pgroonga/pgroonga/commit/0d6f002c24c47da08015af015099c076f0e3f8f0

PR #146, opened 2020-10-16
ci linux: update version to remove.

komainu8:
Because PostgreSQL13 is installed in Ubuntu 20.04 on GitHub Actions.

TODO:
This PR is testing now.
https://github.com/komainu8/pgroonga/actions/runs/309741633

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -64,9 +64,9 @@ jobs:
         run: |
           sudo apt -y -V purge \
             postgresql \
-            postgresql-* \
+            ~n^postgresql- \
             postgresql-client \
-            postgresql-client-*
+            ~n^postgresql-client-
           if [ ${{ matrix.postgresql-version }} -ge 14 ]; then
             sudo sed \
               -i'' \
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -64,14 +64,14 @@ jobs:
         run: |
           sudo apt -y -V purge \
             postgresql \
-            postgresql-12 \
+            postgresql-* \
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -60,16 +60,12 @@ jobs:
...
-            sudo sed \
-              -i'' \
-              -e 's/main/${{ matrix.postgresql-version }}/g' \
-              /etc/apt/sources.list.d/pgdg.list; \
+          if [ "${{ matrix.postgresql-version }}" != "13" ]; then
+            sudo apt -y -V purge \
+              postgresql \
+              postgresql-13 \
+              postgresql-client \
+              postgresql-client-13
```
kou, line 68: Can we use the following here?

```bash
          sudo apt -y -V purge \
            postgresql \
            postgresql-\* \
            postgresql-client \
            postgresql-client-\*
```

.github/workflows/linux.yml
```diff
@@ -60,16 +60,12 @@ jobs:
...
-          sudo apt -y -V purge \
-            postgresql \
-            postgresql-12 \
-            postgresql-client \
-            postgresql-client-12
-          if [ ${{ matrix.postgresql-version }} -ge 13 ]; then \
-            sudo sed \
-              -i'' \
-              -e 's/main/${{ matrix.postgresql-version }}/g' \
-              /etc/apt/sources.list.d/pgdg.list; \
```
kou, line 72: I want to keep this logic for PostgreSQL 14.

komainu8:
@kou Thank you for your comments. I've fixed them.

PR #148, opened 2020-11-09
appveyor: PostgreSQL 9.6 and 10 binaries use Visual Studio 2015

komainu8:
The CI is timeout on Visual Studio 2013.
because build for libgroonga is slow on it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,19 +2,19 @@ version: "{build}"
 clone_depth: 10
 environment:
   matrix:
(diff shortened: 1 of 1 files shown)


kou:
Does this build PGroonga with Visual Studio 2015 with PostgreSQL built with Visual Studio 2013?
Does the built binary work?

komainu8:
I see.
Probably, the built binary doesn't work.

PR #166, opened 2020-12-22
test: follow Groonga 10.1.0 change

komainu8:
A line feed was excepted from the target of highlight by the following commit.
https://github.com/groonga/groonga/commit/8a0c7ee3218a8e342c079c56c37a6b09e284368b

This PR should merge after Groonga 10.1.0 will be released.
Because if we merge this now, tests fail except "PostgreSQL 13 + Visual Studio 2019 + Groonga master".


kou:
#165 should be reverted.

kou:
It seems that expected values are wrong.

PR #196, opened 2021-11-07
Fix a crash bug when we execute EXPLAIN ANALYZE

komainu8:
Currently, if we execute EXPLAIN ANALYZE as below, PGroonga crashes.

```sql
CREATE TABLE memos (
  content text
);

INSERT INTO memos VALUES ('PostgreSQL is a RDBMS.');

EXPLAIN ANALYZE SELECT * FROM memos WHERE content &@~ 'PostgreSQL';
```

### DIFF
--- a/src/pgrn-row-level-security.c
+++ b/src/pgrn-row-level-security.c
@@ -101,6 +101,10 @@ PGrnCheckRLSEnabledSeqScan(FunctionCallInfo fcinfo)
 	if (!portal) {
 		return false;
 	}
+	if (!(portal->queryDesc)) {/* For EXPLAIN ANALYZE */
+		/* If portal->queryDesc is NULL, PostgreSQL's Executor is not active. */
+		return false;
+	}
 	econtext = PGrnFindTargetExprContext(portal->queryDesc->planstate, fcinfo);
 	if (!econtext) {
 		/* For safety */
--- a/sql/full-text-search/jsonb/match-v2/row-level-security/explain-analyze.sql
+++ b/sql/full-text-search/jsonb/match-v2/row-level-security/explain-analyze.sql
@@ -1,0 +1,19 @@
+CREATE TABLE memos (
+  record jsonb
+);
+
+INSERT INTO memos VALUES
+  ('{"title": "PostgreSQL", "content": "PostgreSQL is a RDBMS."}');
+
+SET enable_seqscan = on;
+SET enable_indexscan = off;
+SET enable_bitmapscan = off;
+
+EXPLAIN ANALYZE
+SELECT record
+  FROM memos
+ WHERE record &@ 'groonga'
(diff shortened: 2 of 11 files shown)


kou:
Is it really caused by `EXPLAIN ANALYZE`?
Existing tests that use `EXPLAIN` such as https://github.com/pgroonga/pgroonga/blob/master/sql/full-text-search/text/single/query-v2/row-level-security/indexscan.sql#L29 don't need this. And replcaing `EXPLAIN (COSTS OFF)` with `EXPLAIN ANALYZE` don't cause a crash:

```diff
```

kou:
It seems that "no index" and "`EXPLAIN ANALYZE`" causes this:

```diff
diff --git a/sql/full-text-search/text/single/query-v2/row-level-security/seqscan.sql b/sql/full-text-search/text/single/query-v2/row-level-security/seqscan.sql
index 0fd3c87..ef44696 100644
--- a/sql/full-text-search/text/single/query-v2/row-level-security/seqscan.sql
+++ b/sql/full-text-search/text/single/query-v2/row-level-security/seqscan.sql
@@ -23,7 +23,7 @@ SET enable_bitmapscan = off;
 
 SET SESSION AUTHORIZATION alice;
 \pset format unaligned
-EXPLAIN (COSTS OFF)
+EXPLAIN ANALYZE
 SELECT id, content
   FROM memos
  WHERE content &@~ 'rdbms OR engine'
```

BTW, we should add a test for a fix.

src/pgrn-row-level-security.c
```diff
@@ -101,6 +101,10 @@ PGrnCheckRLSEnabledSeqScan(FunctionCallInfo fcinfo)
 	if (!portal) {
 		return false;
 	}
+	if (!(portal->queryDesc)) {/* For EXPLAIN ANALYZE */
+		/* If portal->queryDesc is NULL, PostgreSQL's Executor is not active. */
+		return false;
```
kou, line 106: Can we return `false` here?
PGroonga functions may show arguments in error message only when `EXPLAIN ANALYZE`?

komainu8:
> It seems that "no index" and "`EXPLAIN ANALYZE`" causes this:

You are right.
This problem occure when PotgreSQL select sequential scan and we execute `EXPLAIN ANALYZE`.

src/pgrn-row-level-security.c
komainu8: I have already returned false here.

kou: I wanted to say that `false` may not be good here.
``EXPLAIN ANALYZE SELECT * FROM table WHERE column &` 'xxx'`` will show the given argument in error message with `false`.

kou:
Are you adding tests for all functions...?
We need just one case for this because there are tests that confirm RLS check is used.

kou:
Fixed on master.

PR #200, opened 2021-12-17
packages yum: drop support for CentOS 8

komainu8:
Because CentOS 8 will reach EOL at 2021-12-31.


kou:
Could you also remove `packages/*/yum/centos-8/`?

komainu8:
Ah...
I'm sorry. I overlooked those directories.
I've removed those directories.

PR #205, opened 2022-02-12
ci linux: Remove accidental quotes

andersk:
Apparently APT accepted these, but that’s not documented so we shouldn’t rely on it.


kou:
Good catch!

PR #255, opened 2022-11-16
Add dynamic snippet length to pgroonga_snippet_html function

askdkc:
GitHub: fix GH-253 

New syntax:

```text
text[] pgroonga_snippet_html(target text,
                             ARRAY[keyword1 text, keyword2 text, ...],
                             width integer DEFAULT 200)
```

`width` is an optional argument.
The default value of `width` is 200.
If we specify a value less than a keyword length to `width`, no records are output.

Sample query:

```sql
SELECT unnest(pgroonga_snippet_html(
  'Groonga is a fast and accurate full text search engine based on ' ||
  'inverted index. One of the characteristics of Groonga is that a ' ||
  'newly registered document instantly appears in search results. ' ||
  'Also, Groonga allows updates without read locks. These characteristics ' ||
  'result in superior performance on real-time applications.' ||
  E'\n' ||
  E'\n' ||
  'Groonga is also a column-oriented database management system (DBMS). ' ||
  'Compared with well-known row-oriented systems, such as MySQL and ' ||
  'PostgreSQL, column-oriented systems are more suited for aggregate ' ||
  'queries. Due to this advantage, Groonga can cover weakness of ' ||
  'row-oriented systems.',
  ARRAY['Groonga'],
  8));
--                 unnest                 
-- ---------------------------------------
--  <span class="keyword">Groonga</span> 
--   <span class="keyword">Groonga</span>
--   <span class="keyword">Groonga</span>
-- (3 rows)
```


HashidaTKS:
@kou 

HashidaTKS:
@askdkc 

Pull Request ありがとうございます！

頂いた内容で、長さの取得に `PG_GETARG_ARRAYTYPE_P` ではなく `PG_GETARG_INT32` （整数型で取得） を使えばほぼほぼ動いている感じでした！

sql/function/snippet-html/width/equal_to_keyword.sql
```diff
@@ -0,0 +1,14 @@
...
+  'Also, Groonga allows updates without read locks. These characteristics ' ||
+  'result in superior performance on real-time applications.' ||
+  E'\n' ||
+  E'\n' ||
+  'Groonga is also a column-oriented database management system (DBMS). ' ||
+  'Compared with well-known row-oriented systems, such as MySQL and ' ||
+  'PostgreSQL, column-oriented systems are more suited for aggregate ' ||
+  'queries. Due to this advantage, Groonga can cover weakness of ' ||
+  'row-oriented systems.',
+  ARRAY['Groonga'], 7));
```
kou, line 14: Could you use a separated line for `width` for readability?

```suggestion
  ARRAY['Groonga'],
  7));
```

src/pgrn-snippet-html.c
```diff
@@ -120,18 +119,19 @@ PGrnSnipExec(grn_obj *snip, text *target, ArrayType **snippetArray)
 }
 
 /**
- * pgroonga.snippet_html(target text, keywords text[]) : text[]
+ * pgroonga_snippet_html(target text, keywords text[], unsigned int width) : text[]
```
kou, line 122:
```suggestion
 * pgroonga_snippet_html(target text, keywords text[], width integer DEFAULT 200) : text[]
```

src/pgrn-snippet-html.c
```diff
@@ -120,18 +119,19 @@ PGrnSnipExec(grn_obj *snip, text *target, ArrayType **snippetArray)
...
  */
 Datum
 pgroonga_snippet_html(PG_FUNCTION_ARGS)
 {
 	const char *tag = "[snippet-html]";
 	text *target = PG_GETARG_TEXT_PP(0);
 	ArrayType *keywords = PG_GETARG_ARRAYTYPE_P(1);
+	unsigned int width = PG_GETARG_INT32(2);
```
kou, line 130: Could you add a check for `PG_GETARG_INT32(2) >= 0`?
We can use `ereport(ERROR, (errcoe(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("...")))` to report an error for invalid input.

kou (commented):
Could you use `-` not `_` for word separator in file name?
Could you also add a test for `width => N` (mixed notation https://www.postgresql.org/docs/current/sql-syntax-calling-funcs.html#SQL-SYNTAX-CALLING-FUNCS-MIXED ) usage?

src/pgrn-snippet-html.c
HashidaTKS: Added

sql/function/snippet-html/width/equal_to_keyword.sql
HashidaTKS: Fixed.

HashidaTKS:
@kou 

I have addressed your comments.
Would you re-review this PR?

src/pgrn-snippet-html.c
```diff
@@ -120,18 +119,28 @@ PGrnSnipExec(grn_obj *snip, text *target, ArrayType **snippetArray)
...
+	int width = PG_GETARG_INT32(2);
 	grn_obj *snip;
 	ArrayType *snippets;
 
-	snip = PGrnSnipCreate(keywords, tag);
+	if (width <= 0)
+	{
+		ereport(ERROR,
+		        (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+		         errmsg("%s width must be a positive number: %d",
```
kou, line 138:
```suggestion
		         errmsg("pgroonga: %s width must be a positive number: %d",
```

kou:
Thanks!

PR #260, opened 2022-11-21
ci: upload source archive

HashidaTKS:
Added a `Source` job for CI that uploads a source archive.


HashidaTKS:
@kou  @komainu8 

.github/workflows/package.yml
```diff
@@ -10,6 +10,45 @@ on:
...
+            qemu-user-static \
+            ruby
+          sudo gem install bundler
+          bundle install
+      - name: Clone dependencies
+        run: |
+          cd ..
+          git clone --depth 1 https://github.com/groonga/groonga.git
+          cd groonga
+          git submodule update --init --force --recursive --depth 1
```
kou, line 33: Could you use the latest released archive instead?

.github/workflows/package.yml
```diff
@@ -10,6 +10,45 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt -y -V install \
+            qemu-user-static \
```
kou, line 24: Do we need this?
We don't use Docker image for arm64.

.github/workflows/package.yml
```diff
@@ -10,6 +10,45 @@ on:
...
+          git submodule update --init --force --recursive --depth 1
+      - name: Build archive
+        run: |
+          rake dist
+
+      # Artifact
+      - uses: actions/upload-artifact@v2
+        with:
+          name: source
+          path: pgroonga-*.tar.gz
```
kou, line 42: Could you use this in `linux` job?

komainu8:
Could you add `needs: source` in `linux` job?
Because `linux` job use a source archive that build in `source` job.

.github/workflows/package.yml
HashidaTKS: Oh, is `qemu-user-static` installed for Docker image for arm64 at this context?
Ok, I will remove this.

kou: Yes.

.github/workflows/package.yml
HashidaTKS: In my second thought, the archive file doesn't depend on groonga.
So I have remove this step.

Related to this, the linux job use not the latest released archive but the latest Groonga repository.

https://github.com/pgroonga/pgroonga/blob/master/.github/workflows/package.yml#L88

I will fix it in another PR.

.github/workflows/package.yml
HashidaTKS: Thanks, fixed.

.github/workflows/package.yml
HashidaTKS: Changed to use this archive,

HashidaTKS:
@kou @komainu8 

Thank you, I have addressed your comments.
Please re-review this PR.

Issue #262, opened 2022-11-22
ci packages: use the latest released archive


HashidaTKS:
`linux` job uses not the latest released archive but the latest Groonga repository.

https://github.com/pgroonga/pgroonga/blob/master/.github/workflows/package.yml#L88

I will fix it in another PR.

_Originally posted by @HashidaTKS in https://github.com/pgroonga/pgroonga/pull/260#discussion_r1029001407_

kou:
It seems that it's just used for `packages/pgroonga-package-task.rb`. So we can use the Groonga repository.

PR #264, opened 2022-11-22
ci packages: use latest archive instead of latest repository

HashidaTKS:
Use the latest archive instead of the latest repository.


kou:
We don't need this.
See also: https://github.com/pgroonga/pgroonga/issues/262#issuecomment-1323285085

PR #266, opened 2022-11-24
Don't use REINDEX against a partition table

komainu8:
TODO: Test

Because REINDEX does not support a partition table.


kou:
We can use `REINDEX INDEX` for partitioned index (not partition index, parent's index).

komainu8:
I see. Thank you for your comment.

PR #280, opened 2022-12-28
ci: Use setup-ruby for linux packages

HashidaTKS:


HashidaTKS:
@kou @komainu8 

kenhys (approved):
LGTM.

.github/workflows/package.yml
```diff
@@ -109,13 +109,15 @@ jobs:
...
+          bundler-cache: true
       - name: Install dependencies
         run: |
           sudo apt update
           sudo apt -y -V install \
-            qemu-user-static \
-            ruby
-          sudo gem install bundler
+            qemu-user-static
           bundle install
```
kou, line 121: We don't need this because `ruby/setup-ruby`'s `bundler-cache: true` does it.

```suggestion
```

PR #334, opened 2023-05-11
tools: add a new tool for creating release announce template for Blog

komainu8:
This tool get contens of release in GitHub release page.
This tool outputs yyyy-mm-dd-pgroonga-x.y.z.md.

Usage:

 $ ./create-release-announce-template-for-blog.sh

Currently, this tool only output a blog for English.


otegami:
How about closing this PR because we don't handle it for while?

kou:
If we need this, we should do this in `Rakefile`.

PR #339, opened 2023-05-19
tool: add a newly tool for generating PGroonga upgrade page

komainu8:


kou:
How about putting this to pgroonga.github.io not here?

komainu8:
Oh.
I see. I'll move this modification.

PR #401, opened 2024-02-06
tools/download-rpm: add download targets

komainu8:
We can download RPM for groonga-plugin-suggest and groonga if we use "USE_GROONGA_PLUGIN_SUGGEST=yes" and "USE_GROONGA_BIN=yes" as below.

```bash
USE_GROONGA_PLUGIN_SUGGEST=yes \
USE_GROONGA_BIN=yes \
  /pgroonga/tools/download-rpm.sh \
    ${PGROONGA_VERSION} \
    ${GROONGA_VERSION} \
    ${POSTGRESQL_VERSION}
```

### DIFF
--- a/tools/download-rpm.sh
+++ b/tools/download-rpm.sh
@@ -115,3 +115,13 @@ if [ ${USE_MECAB:-no} = "yes" ]; then
   download_recursive \
     groonga-tokenizer-mecab-${GROONGA_VERSION}-1.el${OS_VERSION}
 fi
+
+if [ ${USE_PLUGIN_SUGGEST:-no} = "yes" ]; then
+  download_recursive \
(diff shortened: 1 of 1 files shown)


tools/download-rpm.sh
```diff
@@ -115,3 +115,13 @@ if [ ${USE_MECAB:-no} = "yes" ]; then
   download_recursive \
     groonga-tokenizer-mecab-${GROONGA_VERSION}-1.el${OS_VERSION}
 fi
+
+if [ ${USE_PLUGIN_SUGGEST:-no} = "yes" ]; then
```
kou, line 119:
```suggestion
if [ ${USE_GROONGA_PLUGIN_SUGGEST:-no} = "yes" ]; then
```

PR #407, opened 2024-02-22
ci: Enable the test that was disabled

abetomo:
Adding the parameters will cause the test to pass. 
https://github.com/pgroonga/pgroonga/discussions/406#discussioncomment-8550910


kou:
We have more places that removes the test file.
Could you also remove the deletion in this PR?

abetomo:
`full-text-search/text/single/declarative-partitioning.sql` is enabled on Windows.
I did not enable the other removed tests because they seemed to fail for the reasons noted in the comments.

kou:
Wow! The passed with these parameters!

PR #430, opened 2024-04-22
higlight-html: fix the highlight position when we specify loose_blank option

otegami:
TODO: write commit message and implementation.


komainu8:
The cause of this problem is Groonga.
Therefore, we open this problem in Groonga repository.

PR #432, opened 2024-05-01
Add pgroonga_list_broken_indexes() skeleton

abetomo:
Prepare to implement a function that lists the indexes of PGroonga's that may be broken.

### DIFF
--- a/data/pgroonga.sql
+++ b/data/pgroonga.sql
@@ -434,6 +434,12 @@ CREATE FUNCTION pgroonga_wal_set_applied_position()
 	LANGUAGE C
 	STRICT;
 
+CREATE FUNCTION pgroonga_list_broken_indexes()
+	RETURNS void /* todo */
+	AS 'MODULE_PATHNAME', 'pgroonga_list_broken_indexes'
+	LANGUAGE C
+	STRICT
+	PARALLEL SAFE;
 
 /* v1 */
 CREATE FUNCTION pgroonga_match_term(target text, term text)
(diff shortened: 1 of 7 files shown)


data/pgroonga.sql
```diff
@@ -434,6 +434,12 @@ CREATE FUNCTION pgroonga_wal_set_applied_position()
 	LANGUAGE C
 	STRICT;
 
+CREATE FUNCTION pgroonga_list_broken_indexes()
+	RETURNS void /* todo */
+	AS 'MODULE_PATHNAME', 'pgroonga_list_broken_indexes'
+	LANGUAGE C
+	STRICT
+	PARALLEL SAFE;
```
kou, line 442: Could you add the same SQL to `data/pgroonga--3.2.0--3.2.1.sql`?
Could you also add `DROP FUNCTION pgroonga_list_broken_indexes;` to `data/pgroonga--3.2.1--3.2.0.sql`?

abetomo: I was going to add it after it was fixed because there will always be changes, but should I add it already?

kou: Yes, please.
FYI: `data/pgroonga--3.2.1--3.2.0.sql` will not be changed.

abetomo: I got it!

src/pgrn-list-broken-indexes.c
```diff
@@ -0,0 +1,17 @@
+#include "pgroonga.h"
+#include <funcapi.h>
```
kou, line 2:
```suggestion

#include <funcapi.h>
```

PR #434, opened 2024-05-06
list-broken-indexes: Returns the list of PGroonga indexes being locked

abetomo:
Prepare to implement a function that lists the indexes of PGroonga's that may be broken. This commit returns a list of indexes being locked. The next step should also return corrupt indexes.


kou:
Could you run `pre-commit run -a`?

src/pgrn-list-broken-indexes.c
```diff
@@ -10,12 +13,84 @@
...
+	bool is_locked = false;
+
+	columns = grn_hash_create(
+		ctx, NULL, sizeof(grn_id), 0, GRN_TABLE_HASH_KEY | GRN_HASH_TINY);
+	if (!columns)
+	{
+		PGrnCheck("failed to create columns container for checking locks: "
+				  "<%s>",
+				  PGrnInspectName(table));
+	}
```
kou, line 37: We don't need this condition because `PGrnCheck()` does nothing if `ctx->rc == GRN_SUCCESS`:

```suggestion
	PGrnCheck("failed to create columns container for checking locks: "
			  "<%s>",
			  PGrnInspectName(table));
```

src/pgrn-list-broken-indexes.c
```diff
@@ -10,12 +13,84 @@
...
 {
 	Relation indexes;
 	TableScanDesc scan;
 } PGrnListBrokenIndexData;
 
+static bool
+PGrnIsLockedColumn(grn_obj *table)
+{
+	grn_hash *columns;
+	bool is_locked = false;
```
kou, line 28:
```suggestion
	bool isLocked = false;
```

src/pgrn-list-broken-indexes.c
```diff
@@ -10,12 +13,84 @@
...
+static grn_ctx *ctx = &PGrnContext;
+
 typedef struct
 {
 	Relation indexes;
 	TableScanDesc scan;
 } PGrnListBrokenIndexData;
 
+static bool
+PGrnIsLockedColumn(grn_obj *table)
```
kou, line 25:
```suggestion
PGrnTableHaveLockedColumn(grn_obj *table)
```

src/pgrn-list-broken-indexes.c
```diff
@@ -62,13 +137,19 @@ pgroonga_list_broken_indexes(PG_FUNCTION_ARGS)
 			continue;
 		}
 
+		if (!PGrnIsLockedSources(index) && !PGrnIsLockedLexicon(index))
```
kou, line 140: How about adding `PGrnIndexIsBroken()` and use it here?

```suggestion
		if (!PGrnIndexIsBroken(index))
```

src/pgrn-list-broken-indexes.c
```diff
@@ -10,12 +13,90 @@
...
+}
+
+static bool
+PGrnIsLockedLexicon(Relation index)
+{
+	grn_obj *lexicon;
+	int i;
+	for (i = 0; i < index->rd_att->natts; i++)
+	{
+		lexicon = PGrnLookupLexicon(index, i, ERROR);
```
kou, line 79:
```suggestion
	int i;
	for (i = 0; i < index->rd_att->natts; i++)
	{
		grn_obj *lexicon = PGrnLookupLexicon(index, i, ERROR);
```

PR #442, opened 2024-05-17
Check the lag with pgroonga_wal_status()

abetomo:
* List PGroonga indexes with unapplied WAL in pgroonga_wal_status()
* List all PGroonga indexes if primary log position and lag

### DIFF
--- a/data/pgroonga.sql
+++ b/data/pgroonga.sql
@@ -435,9 +435,18 @@ CREATE FUNCTION pgroonga_wal_set_applied_position()
 	STRICT;
 
 CREATE FUNCTION pgroonga_list_lagged_indexes()
-	RETURNS void /* todo */
-	AS 'MODULE_PATHNAME', 'pgroonga_list_lagged_indexes'
-	LANGUAGE C
+	RETURNS TABLE (name text) AS $$
+		SELECT name FROM pgroonga_wal_status()
+		WHERE current_block != last_block
+			OR current_offset != current_offset
+			OR current_size != last_size
+		UNION
+		SELECT name FROM pgroonga_wal_status()
+		WHERE EXISTS(
+			SELECT 1 FROM pg_stat_wal_receiver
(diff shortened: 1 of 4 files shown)


abetomo:
If this looks good, revert as https://github.com/pgroonga/pgroonga/commit/c2a01c0d1ee587ee063aede3510d9a9adcbb1b43 is not needed.

data/pgroonga.sql
```diff
@@ -435,9 +435,18 @@ CREATE FUNCTION pgroonga_wal_set_applied_position()
...
-	AS 'MODULE_PATHNAME', 'pgroonga_list_lagged_indexes'
-	LANGUAGE C
+	RETURNS TABLE (name text) AS $$
+		SELECT name FROM pgroonga_wal_status()
+		WHERE current_block != last_block
+			OR current_offset != current_offset
+			OR current_size != last_size
+		UNION
+		SELECT name FROM pgroonga_wal_status()
+		WHERE EXISTS(
```
kou, line 445: Can we use one `SELECT`?

```suggestion
		OR EXISTS(
```

abetomo: Good point.

kou:
We can remove the C implementation.

data/pgroonga.sql
```diff
@@ -434,12 +434,45 @@ CREATE FUNCTION pgroonga_wal_set_applied_position()
...
+					OR current_offset != current_offset
+					OR current_size != last_size
+					OR EXISTS(
+						SELECT 1 FROM pg_stat_wal_receiver
+						WHERE flushed_lsn != latest_end_lsn
+					);
+			' LANGUAGE SQL
+			STRICT
+			PARALLEL SAFE;
+	ELSIF SUBSTR(rec.setting, 1, 2) = '12' THEN
```
kou, line 459: We can use `ELSE` because we don't support PostgreSQL 11 or earlier.

PR #444, opened 2024-05-21
Check with PGrnCheck() after `grn_expr_append_*` at PGrnSearchBuildConditionJSONScript()

abetomo:
If it failed, PGroonga crashes, so it is checked by PGrnCheck().

### DIFF
--- a/src/pgrn-jsonb.c
+++ b/src/pgrn-jsonb.c
@@ -1596,13 +1596,31 @@ PGrnSearchBuildConditionJSONScript(PGrnSearchData *data,
 								   grn_obj *filter,
 								   unsigned int *nthCondition)
 {
+	const char *tag = "[build-condition][json-script]";
+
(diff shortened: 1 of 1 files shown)


src/pgrn-jsonb.c
```diff
@@ -1596,13 +1596,31 @@ PGrnSearchBuildConditionJSONScript(PGrnSearchData *data,
 								   grn_obj *filter,
 								   unsigned int *nthCondition)
 {
+	const char *tag = "[build-condition][json-script]";
+
 	grn_expr_append_obj(ctx, data->expression, subFilter, GRN_OP_PUSH, 1);
+	PGrnCheck("jsonb: %s: failed to append sub_filter", tag);
```
kou, line 1602:
```suggestion
	PGrnCheck("jsonb: %s: failed to append sub_filter()", tag);
```

PR #445, opened 2024-05-21
Check with PGrnCheck() after `grn_expr_append_op` in pgroonga.c

abetomo:
If it failed, PGroonga crashes, so it is checked by PGrnCheck().

This commit does not add checks to all `grn_expr_append_op`.
It is split up because of the many targets.

### DIFF
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -1437,7 +1437,7 @@ PGrnCollectScoreMultiColumnPrimaryKey(Relation table,
 									  HeapTuple tuple,
 									  PGrnScanOpaque so)
 {
-	const char *tag = "pgroonga: [score][multi-column-primary-key][collect]";
+	const char *tag = "[score][multi-column-primary-key][collect]";
 	double score = 0.0;
 	TupleDesc desc;
 	grn_obj *expression;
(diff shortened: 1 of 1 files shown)


src/pgroonga.c
```diff
@@ -1437,6 +1437,7 @@ PGrnCollectScoreMultiColumnPrimaryKey(Relation table,
 									  HeapTuple tuple,
 									  PGrnScanOpaque so)
 {
+	const char *tag = "pgroonga: [score][multi-column-primary-key][collect]";
```
kou, line 1440: `pgroonga: ` is prepended by `PGrnCheck()` automatically.

```suggestion
	const char *tag = "[score][multi-column-primary-key][collect]";
```

src/pgroonga.c
```diff
@@ -5547,20 +5562,33 @@ PGrnSearchBuildConditionLikeMatchFlush(grn_obj *expression,
...
 	grn_expr_append_op(ctx, expression, GRN_OP_MATCH, 2);
+	PGrnCheck("%s: failed to append operator: %s",
+			  tag,
+			  grn_operator_to_string(GRN_OP_MATCH));
 	if (*nKeywords > 0)
+	{
 		grn_expr_append_op(ctx, expression, GRN_OP_OR, 2);
+		PGrnCheck("%s: failed to append operator: %s",
+				  tag,
+				  grn_operator_to_string(GRN_OP_MATCH));
```
kou, line 5590:
```suggestion
				  grn_operator_to_string(GRN_OP_OR));
```

PR #447, opened 2024-05-21
list-broken-indexes: Add PGrnIsBrokenGrnObj()

abetomo:
Refactoring:
Implemented a function for common checks.

### DIFF
--- a/src/pgrn-list-broken-indexes.c
+++ b/src/pgrn-list-broken-indexes.c
@@ -59,18 +59,23 @@ PGrnTableHaveBrokenColumn(grn_obj *table)
 }
 
 static bool
+PGrnIsBrokenGrnObj(grn_obj *obj)
+{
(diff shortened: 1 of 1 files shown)


src/pgrn-list-broken-indexes.c
```diff
@@ -59,18 +59,23 @@ PGrnTableHaveBrokenColumn(grn_obj *table)
 }
 
 static bool
-PGrnIsBrokenSources(Relation index)
+PGrnIsBrokenGrnObj(grn_obj *obj)
```
kou, line 62: We call `PGrnTableHaveBrokenColumn()`. So we should use `Table` not `GrnObj` here. `GrnObj` is too generic because a column is also a `grn_obj`.

```suggestion
PGrnIsBrokenTable(grn_obj *table)
```

kou:
Oh... I forgot to fix the title and the commit message...

PR #454, opened 2024-05-22
list-broken-indexes: Support jsonb

abetomo:


sql/function/list-broken-indexes/jsonb/locked.sql
```diff
@@ -0,0 +1,19 @@
...
+
+INSERT INTO logs VALUES (1, '{"message": "Hello World"}');
+INSERT INTO logs VALUES (2, '{"message": "This is a pen"}');
+INSERT INTO logs VALUES (3, '{"message": "Good-by World"}');
+CREATE INDEX pgroonga_index ON logs
+  USING pgroonga (record pgroonga_jsonb_ops_v2);
+
+select pgroonga_command(
+  'lock_acquire',
+  ARRAY['target_name', replace(pgroonga_table_name('pgroonga_index'), 'Sources', 'JSONPaths') || '_0']
```
abetomo, line 14:
```sql
select replace(pgroonga_table_name('pgroonga_index'), 'Sources', 'JSONPaths') || '_0';
     ?column?
-------------------
 JSONPaths203941_0
(1 row)
```

What is a better way to get table names for jsonb?

kou: We can use this because we don't have shorter way than this.

src/pgrn-list-broken-indexes.c
```diff
@@ -93,10 +94,83 @@ PGrnIsBrokenLexicon(Relation index)
...
+	if (PGrnIsBrokenSources(index))
+	{
+		return true;
+	}
+
+	desc = RelationGetDescr(index);
+	for (i = 0; i < desc->natts; i++)
+	{
+		Form_pg_attribute attribute = TupleDescAttr(desc, i);
+		if (PGrnAttributeIsJSONB(attribute->atttypid))
```
kou, line 160: We can simplify `jsonb` check like:

https://github.com/pgroonga/pgroonga/blob/52e7d2ed997b2e32760a6163083d9559e89115f5/src/pgroonga.c#L4723-L4724

Because we don't support multi-column `jsonb` index.

PR #455, opened 2024-05-22
Add PGrnExprAppendConstString()

abetomo:
Check with PGrnCheck() after grn_expr_append_const_str().

### DIFF
--- a/src/pgrn-groonga.c
+++ b/src/pgrn-groonga.c
@@ -695,18 +695,18 @@ PGrnExprAppendOp(grn_obj *expr,
 }
 
 void
-PGrnExprAppendConstStr(grn_obj *expr,
-					   const char *str,
-					   unsigned int strSize,
-					   grn_operator op,
-					   int nArgs,
-					   const char *tag)
+PGrnExprAppendConstString(grn_obj *expr,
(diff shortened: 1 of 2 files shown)


src/pgrn-groonga.c
```diff
@@ -693,3 +693,20 @@ PGrnExprAppendOp(grn_obj *expr,
 				  nArgs);
 	}
 }
+
+void
+PGrnExprAppendConstStr(grn_obj *expr,
+					   const char *str,
+					   unsigned int strSize,
```
kou, line 700:
```suggestion
					   const char *string,
					   unsigned int stringSize,
```

src/pgrn-groonga.c
```diff
@@ -693,3 +693,20 @@ PGrnExprAppendOp(grn_obj *expr,
 				  nArgs);
 	}
 }
+
+void
+PGrnExprAppendConstStr(grn_obj *expr,
```
kou, line 698:
```suggestion
PGrnExprAppendConstString(grn_obj *expr,
```

kou (commented):
We don't want to abbreviate as much as possible.

PR #484, opened 2024-06-05
ci: skip upgrade test on the latest code name

otegami:


packages/apt/test.sh
```diff
@@ -152,20 +152,23 @@ echo "::group::Upgrade"
...
-  ${repositories_dir}/${os}/pool/${code_name}/*/*/*/*_${architecture}.deb
-psql upgrade -c 'ALTER EXTENSION pgroonga UPDATE'
+if sudo apt install -V -y ${pgroonga_package}; then
+  # Disable upgrade test for first time packages.
+  # TODO: Automate this. We don't want to maintain this manually.
+  # case ${postgresql_version} in
+  #   16) # TODO: Remove this after 3.1.4 release.
+  #     exit
+  #     ;;
+  # esac
```
otegami, line 162: Is it the same context about the unreleased package?
If so, I will delete this comment.

packages/apt/test.sh
```diff
@@ -152,20 +152,23 @@ echo "::group::Upgrade"
...
-#     ;;
-# esac
-
-sudo apt install -V -y ${pgroonga_package}
-createdb upgrade
-psql upgrade -c 'CREATE EXTENSION pgroonga'
-apt install -V -y \
-  ${repositories_dir}/${os}/pool/${code_name}/*/*/*/*_${architecture}.deb
-psql upgrade -c 'ALTER EXTENSION pgroonga UPDATE'
+if sudo apt install -V -y ${pgroonga_package}; then
```
otegami, line 155: Could you tell me if there is the better way to check whether the package exists or not?
I tried to check the following commands but it didn't work.
- `apt install --dry-run package`: It succeeded using it
- `apt-cache show package`: It succeeded using it

packages/apt/test.sh
kou: Yes. (It's only for the unreleased package.)

packages/apt/test.sh
kou: I think that at least this approach isn't suitable. Because we can't detect why it's failed when it's failed. It may be failed in normal package installation process instead of "the packages doesn't exist".

otegami: fix: [8599cb2](https://github.com/pgroonga/pgroonga/pull/484/commits/8599cb2d69de82f3e394d6e1392f15402ddbee8b)
I've just fixed like that I check whether the package exists or not after purging the info. 
Because I didn't read we had already installed the package in the precious step and got the meta information about the package.

packages/apt/test.sh
```diff
@@ -150,22 +150,18 @@ fi
...
-#     ;;
-# esac
-
-sudo apt install -V -y ${pgroonga_package}
-createdb upgrade
-psql upgrade -c 'CREATE EXTENSION pgroonga'
-apt install -V -y \
-  ${repositories_dir}/${os}/pool/${code_name}/*/*/*/*_${architecture}.deb
-psql upgrade -c 'ALTER EXTENSION pgroonga UPDATE'
+sudo apt-get purge -V -y ${pgroonga_package}
```
kou, line 153: 他のところで`apt`を使っているので`apt`でいいっす。

```suggestion
sudo apt purge -V -y ${pgroonga_package}
```

otegami: fix: [6a32183](https://github.com/pgroonga/pgroonga/pull/484/commits/6a321834d69f13358730cdcfc0c41246f9e12784)
他の部分と合わせて、`apt` を使う方針に修正しました。

packages/apt/test.sh
```diff
@@ -152,20 +152,16 @@ echo "::group::Upgrade"
 
 sudo apt purge -V -y ${pgroonga_package}
```
kou, line 153:
```suggestion
apt purge -V -y ${pgroonga_package}
```

packages/apt/test.sh
```diff
@@ -152,20 +152,16 @@ echo "::group::Upgrade"
...
-# esac
-
-sudo apt install -V -y ${pgroonga_package}
-createdb upgrade
-psql upgrade -c 'CREATE EXTENSION pgroonga'
-apt install -V -y \
-  ${repositories_dir}/${os}/pool/${code_name}/*/*/*/*_${architecture}.deb
-psql upgrade -c 'ALTER EXTENSION pgroonga UPDATE'
+if sudo apt show ${pgroonga_package} > /dev/null 2>&1; then
+  sudo apt install -V -y ${pgroonga_package}
```
kou, line 156:
```suggestion
if apt show ${pgroonga_package} > /dev/null 2>&1; then
  apt install -V -y ${pgroonga_package}
```

PR #491, opened 2024-06-18
pgroonga-primary-maintainer: Add a parameter that can limit run hours

abetomo:
Allows it to run only at specific hours to avoid load.

### DIFF
--- a/src/pgroonga-primary-maintainer.c
+++ b/src/pgroonga-primary-maintainer.c
@@ -282,7 +282,7 @@ static bool
 ExecutionHours()
 {
 	pg_time_t now = (pg_time_t) time(NULL);
-	struct pg_tm *tm = pg_localtime(&now, log_timezone);
+	struct pg_tm *tm = pg_localtime(&now, session_timezone);
 
 	if (PGroongaPrimaryMaintainerHoursString == NULL ||
 		PGroongaPrimaryMaintainerHoursString[0] == '\0')
--- a/src/pgroonga-primary-maintainer.c
+++ b/src/pgroonga-primary-maintainer.c
@@ -15,6 +15,7 @@
 #include <storage/latch.h>
 #include <utils/guc.h>
 #include <utils/snapmgr.h>
+#include <utils/varlena.h>
 
 PG_MODULE_MAGIC;
 
@@ -25,12 +26,15 @@ extern PGDLLEXPORT void pgroonga_primary_maintainer_main(Datum datum)
 	pg_attribute_noreturn();
 
 #define TAG "pgroonga: primary-maintainer"
+#define HOURS_SIZE 24
 
 static volatile sig_atomic_t PGroongaPrimaryMaintainerGotSIGTERM = false;
 static volatile sig_atomic_t PGroongaPrimaryMaintainerGotSIGHUP = false;
 static int PGroongaPrimaryMaintainerNaptime = 60;
 static int PGroongaPrimaryMaintainerReindexThreshold =
 	(1024 * 1024 * 1024) / BLCKSZ; // 1GB in size
+static char *PGroongaPrimaryMaintainerHoursString = NULL;
+static bool *PGroongaPrimaryMaintainerHours = NULL;
 static const char *PGroongaPrimaryMaintainerLibraryName =
 	"pgroonga_primary_maintainer";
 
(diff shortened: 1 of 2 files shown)


src/pgroonga-primary-maintainer.c
```diff
@@ -209,6 +213,83 @@ pgroonga_primary_maintainer_wal_size_check_all(void)
 	pgstat_report_activity(STATE_IDLE, NULL);
 }
 
+static bool
+checkPGroongaPrimaryMaintainerHours(char **newval,
```
kou, line 217:
```suggestion
PGroongaPrimaryMaintainerHoursCheck(char **newValue,
```

src/pgroonga-primary-maintainer.c
```diff
@@ -209,6 +213,83 @@ pgroonga_primary_maintainer_wal_size_check_all(void)
...
+		return true;
+
+	rawstring = pstrdup(*newval);
+	if (!SplitDirectoriesString(rawstring, ',', &elemlist))
+	{
+		list_free_deep(elemlist);
+		pfree(rawstring);
+		ereport(LOG,
+				(errcode(ERRCODE_SYNTAX_ERROR),
+				 errmsg("invalid syntax in pgroonga_primary_maintainer.hour")));
```
kou, line 242: Is this correct?
It seems that we need to call `GUC_check_err*()` instead:
* https://github.com/postgres/postgres/blob/c61c0cb3a99285db36806b592bd8c540c98757e5/src/include/utils/guc.h#L443-L453
* https://github.com/postgres/postgres/blob/c61c0cb3a99285db36806b592bd8c540c98757e5/src/backend/replication/syncrep.c#L1009-L1014

src/pgroonga-primary-maintainer.c
```diff
@@ -209,6 +213,83 @@ pgroonga_primary_maintainer_wal_size_check_all(void)
...
+	for (i = 0; i < HOURS_SIZE; i++)
+	{
+		hours[i] = false;
+	}
+
+	if (*newval == NULL || *newval[0] == '\0')
+		return true;
+
+	rawstring = pstrdup(*newval);
+	if (!SplitDirectoriesString(rawstring, ',', &elemlist))
```
kou, line 236: Hmm...
Can we generalize `PGrnOptionParseNames()` and use it instead?
We don't want to allocate memory for this.

src/pgroonga-primary-maintainer.c
```diff
@@ -209,6 +213,83 @@ pgroonga_primary_maintainer_wal_size_check_all(void)
...
+		hours[hour] = true;
+	}
+
+	list_free_deep(elemlist);
+	pfree(rawstring);
+
+	// for 16 or higher
+	// *extra = guc_malloc(LOG, HOURS_SIZE * sizeof(bool));
+	*extra = malloc(HOURS_SIZE * sizeof(bool));
+	memcpy(*extra, hours, HOURS_SIZE * sizeof(bool));
```
kou, line 271: We don't want to allocate this.
Can we define `PGroongaPrimaryMaintainerHours` as `static bool PGroongaPrimaryMaintainerHours[HOURS_SIZE]` instead?
We can re-parse the user input in the assign function.

src/pgroonga-primary-maintainer.c
```diff
@@ -25,12 +26,15 @@ extern PGDLLEXPORT void pgroonga_primary_maintainer_main(Datum datum)
 	pg_attribute_noreturn();
 
 #define TAG "pgroonga: primary-maintainer"
+#define HOURS_SIZE 24
```
kou, line 29:
```suggestion
#define PGRN_PRIMARY_MAINTAINER_HOURS_SIZE 24
```

src/pgroonga-primary-maintainer.c
```diff
@@ -209,6 +213,83 @@ pgroonga_primary_maintainer_wal_size_check_all(void)
...
+
+	// for 16 or higher
+	// *extra = guc_malloc(LOG, HOURS_SIZE * sizeof(bool));
+	*extra = malloc(HOURS_SIZE * sizeof(bool));
+	memcpy(*extra, hours, HOURS_SIZE * sizeof(bool));
+	return true;
+}
+
+static void
+assignPGroongaPrimaryMaintainerHours(const char *newval, void *extra)
```
kou, line 276:
```suggestion
PGroongaPrimaryMaintainerHoursAssign(const char *newValue, void *extra)
```

src/pgroonga-primary-maintainer.c
```diff
@@ -209,6 +213,83 @@ pgroonga_primary_maintainer_wal_size_check_all(void)
...
+{
+	PGroongaPrimaryMaintainerHours = (bool *) extra;
+}
+
+static bool
+ExecutionHours()
```
kou, line 282:
```suggestion
pgroonga_primary_maintainer_is_executable()
```

src/pgroonga-primary-maintainer.c
```diff
@@ -220,6 +301,7 @@ pgroonga_primary_maintainer_main(Datum arg)
 	elog(LOG,
 		 TAG ": reindex_threshold=%d",
 		 PGroongaPrimaryMaintainerReindexThreshold);
+	elog(LOG, TAG ": hours=%s", PGroongaPrimaryMaintainerHoursString);
```
kou, line 304: Is this safe for `NULL`?

abetomo:
Closes `pgroonga-primary-maintainer` implementation because it will stop.

PR #500, opened 2024-06-27
pgroonga-primary-maintainer: Remove unnecessary output

abetomo:
Unnecessary errors are displayed, so stderr also redirects to `/dev/null`.

### DIFF
--- a/test/test-pgroonga-primary-maintainer.rb
+++ b/test/test-pgroonga-primary-maintainer.rb
@@ -16,6 +16,14 @@ class PGroongaPrimaryMaintainerTestCase < Test::Unit::TestCase
     run_command(env, *commane_line)
   end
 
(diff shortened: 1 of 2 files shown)


test/test-pgroonga-primary-maintainer.rb
```diff
@@ -16,6 +16,14 @@ def run_primary_maintainer_command(*options)
     run_command(env, *commane_line)
   end
 
+  def which(command)
+    ENV["PATH"].split(File::PATH_SEPARATOR).each do |path|
+      next unless File.absolute_path?(path)
+      command = File.join(path, command)
+      return command if File.executable?(command)
+    end
```
kou, line 24:
```suggestion
    end
    nil
```

PR #505, opened 2024-06-28
pgroonga-generate-primary-maintainer-service: Add the service file generator to the installation target

abetomo:
GH-498


kou:
Should we change path in tests?
https://github.com/pgroonga/pgroonga/blob/2a733abffbbfc20414a66a154e553ede0d8891c1/test/test-tools-service-file-generator.rb#L13

PR #511, opened 2024-07-02
ci package: prepare Incus to execute tests on it instead of Docker

otegami:
GitHub: ref GH-507

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -248,8 +248,6 @@ jobs:
 
       # Test
       - name: Install Incus
-        if: |
-          startsWith(env.OS, 'almalinux-')
         run: |
           # We can use the official Ubuntu APT repository when
           # ubuntu-latest is Ubuntu 24.04.
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -247,6 +247,25 @@ jobs:
           GH_TOKEN: ${{ github.token }}
 
       # Test
+      - name: Install Incus
+        if: |
+          startsWith(env.OS, 'almalinux-')
+        run: |
+          # We can use the official Ubuntu APT repository when
+          # ubuntu-latest is Ubuntu 24.04.
+          sudo curl -fsSL https://pkgs.zabbly.com/key.asc -o /etc/apt/keyrings/zabbly.asc
+          cat <<SOURCES | sudo tee /etc/apt/sources.list.d/zabbly-incus-stable.sources
+          Enabled: yes
+          Types: deb
+          URIs: https://pkgs.zabbly.com/incus/stable
+          Suites: $(. /etc/os-release && echo ${VERSION_CODENAME})
+          Components: main
+          Architectures: $(dpkg --print-architecture)
+          Signed-By: /etc/apt/keyrings/zabbly.asc
+          SOURCES
+
+          sudo apt update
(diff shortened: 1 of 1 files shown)


.github/workflows/package.yml
```diff
@@ -247,6 +247,25 @@ jobs:
           GH_TOKEN: ${{ github.token }}
 
       # Test
+      - name: Install Incus
+        if: |
+          startsWith(env.OS, 'almalinux-')
```
otegami, line 252: Use Incus on only AlamaLinux at first because we want to test it on this OS.

.github/workflows/package.yml
```diff
@@ -247,6 +247,25 @@ jobs:
...
+        run: |
+          # We can use the official Ubuntu APT repository when
+          # ubuntu-latest is Ubuntu 24.04.
```
otegami, line 255: Actually, we can use `incus` on Ubuntu 24:04 beta but there is a following problem now.
So at first, I want to use `ubuntu-latest` now because it's hard for me figure it out.

ref: https://github.com/pgroonga/pgroonga/issues/507#issuecomment-2199634563

.github/workflows/package.yml
abetomo: Since it doesn't seem hard to migrate to testing on Incus for all OS, could you target all OS?

otegami: Sure.

otegami: fix: [c3fd1e0](https://github.com/pgroonga/pgroonga/pull/511/commits/c3fd1e04436e357c03431c13c588a1867d90fb5a)

abetomo (approved):

.github/workflows/package.yml
kou: Could you fix the comment?

otegami: fix: [41f3ff9](https://github.com/pgroonga/pgroonga/pull/511/commits/41f3ff9a9452f7149c25322e748a953e088be6a7 Sure, I've just updated the comment.

.github/workflows/package.yml
```diff
@@ -247,6 +247,25 @@ jobs:
           GH_TOKEN: ${{ github.token }}
 
       # Test
+      - name: Install Incus
+        run: |
+          # Ubuntu 24.04 beta has already been released as a GitHub Runner image and provides incus.
```
kou, line 252:
```suggestion
          # Ubuntu 24.04 beta has already been released as a GitHub Runner image and provides Incus.
```

.github/workflows/package.yml
```diff
@@ -247,6 +247,25 @@ jobs:
...
+          # However, there is a issue with the combination of Docker, arm64 image and libc-bin on it.
```
kou, line 253:
```suggestion
          # However, there is an issue with the combination of Docker, arm64 image and libc-bin on it.
```

.github/workflows/package.yml
```diff
@@ -247,6 +247,25 @@ jobs:
...
+          # After this issue is resolved, we will use the Ubuntu 24.04 runner image and install incus using apt.
```
kou, line 254:
```suggestion
          # After this issue is resolved, we will use the Ubuntu 24.04 runner image and install Incus using apt.
```

PR #525, opened 2024-08-05
Remove unnecessary pgroonga_condition() definitions

abetomo:
If you upgrade to 3.2.1 with PGroonga already installed, you get the following error.

```
WHERE content &@~ pgroonga_condition('PGroonga');
                           ^
HINT:  Could not choose a best candidate function. You might need to add explicit type casts.
```

The log also reports the following error.

```
ERROR:  function pgroonga_condition(unknown) is not unique at character 84
```

This is due to a bug in `data/pgroonga--3.2.0--3.2.1.sql` where two pgroonga_condition() were defined.
Therefore, the unnecessary pgroonga_condition() definition was removed.

### DIFF
--- a/packages/apt/test.sh
+++ b/packages/apt/test.sh
@@ -2,7 +2,7 @@
 
 set -eux
 
-function run_test () {
+function run_test() {
   echo "::group::Run test"
 
   set +e
--- a/packages/apt/test.sh
+++ b/packages/apt/test.sh
@@ -158,13 +158,14 @@ if apt show ${pgroonga_package} > /dev/null 2>&1; then
   apt install -V -y \
     ${repositories_dir}/${os}/pool/${code_name}/*/*/*/*_${architecture}.deb
   psql upgrade -c 'ALTER EXTENSION pgroonga UPDATE'
+
+  run_test
 else
   echo "Skip because ${pgroonga_package} hasn't been released yet."
 fi
 
 echo "::endgroup::"
 
-run_test
 
 echo "::group::Postpare"
 
--- a/packages/apt/test.sh
+++ b/packages/apt/test.sh
@@ -2,6 +2,33 @@
 
 set -eux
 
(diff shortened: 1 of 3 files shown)


packages/yum/test.sh
```diff
@@ -2,6 +2,33 @@
 
 set -eux
 
+function run_test () {
```
kou:
```suggestion
function run_test() {
```

packages/apt/test.sh
```diff
@@ -118,33 +145,7 @@ pg_regress=$(dirname $(pg_config --pgxs))/../test/regress/pg_regress
...
-  mkdir -p /host-rw/logs
-  cp -a regression.diffs /host-rw/logs/
-  cp -a ${data_dir}/log /host-rw/logs/postgresql || :
-  mkdir -p /host-rw/logs/pgroonga/ || :
-  cp -a ${data_dir}/pgroonga.log* /host-rw/logs/pgroonga/ || :
-  chmod -R go+rx /host-rw/logs/
-  exit ${pg_regress_status}
-fi
-
+run_test
```
otegami, line 148: I think we don't have to run tests if we don't upgrade it.
How about running test in the only upgrading condition?

packages/apt/test.sh
```diff
@@ -163,6 +164,7 @@ fi
 
 echo "::endgroup::"
 
+run_test
```
otegami, line 167: I think we don't have to run tests if we don't upgrade it.
How about running test in the only upgrading condition?

otegami: I mean moving this line to the line160.

abetomo: Woops. i made a mistake in intending to modify it that way.

otegami (approved):
LGTM 👍🏾

komainu8 (approved):

PR #569, opened 2024-10-22
ci: Run all tests, on a periodic run by schedule

abetomo:
Time-expensive tests are skipped on CI, but this may delay noticing failures.
Change to run all tests once a day.
It usually takes about 2 hours, so we set a timeout of 150 minutes.

### DIFF
--- a/test/test-pgroonga-check.rb
+++ b/test/test-pgroonga-check.rb
@@ -151,7 +151,7 @@ SELECT * FROM memos WHERE content %% 'PGroonga';
 
   sub_test_case "list-broken-indexes" do
     test "corrupt" do
-      omit("This test takes over 10 minutes, so skip it.") if ENV["CI"] == "true"
+      omit("This test takes over 10 minutes.") if ENV["CI"] == "true"
 
       run_sql("CREATE TABLE memos (content text);")
       run_sql("CREATE INDEX memos_content ON memos USING pgroonga (content);")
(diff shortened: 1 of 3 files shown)


test/test-pgroonga-check.rb
```diff
@@ -167,7 +167,7 @@ class PGroongaCheckTestCase < Test::Unit::TestCase
         input.puts(<<-"SQL")
 SELECT path
 FROM (
-  SELECT value->1 name, value->2 path
+  SELECT value->1 AS name, value->2 AS path
```
abetomo, line 170: Add `AS`, since it fails in PostgreSQL 12 and 13.

otegami (approved):
LGTM 👍🏾

test/test-pgroonga-check.rb
```diff
@@ -151,7 +151,7 @@ class PGroongaCheckTestCase < Test::Unit::TestCase
 
   sub_test_case "list-broken-indexes" do
     test "corrupt" do
-      omit("This test takes over 10 minutes, so skip it.") if ENV["CI"]
+      omit("This test takes over 10 minutes, so skip it.") if ENV["CI"] == "true"
```
kou, line 154:
```suggestion
      omit("This test takes over 10 minutes.") if ENV["CI"] == "true"
```

kou:
Let's try.

PR #618, opened 2024-11-28
Use PG_RETURN_BOOL() not true/false for Datum

komainu8:


otegami (approved):
👍🏾 
- ref: https://github.com/pgroonga/pgroonga/pull/593#discussion_r1861380560

Issue #607, opened 2024-11-27
Drop support for PostgreSQL 12


otegami:
### What is your problem?

Because it reached EOL.
See: https://www.postgresql.org/support/versioning/

otegami:
What we will do

- [x] Drop support for PostgreSQL 12 in CI
- Drop support for PostgreSQL 12 from source
  - Remove unnecessary macros
    - [x] PGRN_SUPPORT_OPTION_LOCK_MODE
    - [x] PGRN_HAVE_BUILD_RELOPTIONS
    - [ ] PGRN_HAVE_JSONB_DATETIME
    - [ ] PGRN_INDEX_BUILD_CALLBACK_USE_ITEM_POINTER
    - [ ] PGRN_WL_EXIT_ON_PM_DEATH WL_EXIT_ON_PM_DEATH
    - [ ] PGRN_HAVE_COMMON_HASHFN_H
    - [ ] PGRN_INDEX_AM_ROUTINE_HAVE_AM_USE_MAINTENANCE_WORK_MEM
    - [ ] PGRN_INDEX_AM_ROUTINE_HAVE_AM_PARALLEL_VACUUM_OPTIONS

otegami:
We've dropped the support for PostgreSQL 12.

PR #620, opened 2024-11-28
 test query-expand: drop support for PostgreSQL 12

otegami:
GitHub: GH-607

We dropped support for PostgreSQL 12.
So we don't have to consider the following case to use PostgreSQL 12.
- ref: https://github.com/pgroonga/pgroonga/commit/b5bba12cbbb3dbcbb3d5f2c50e434b48a43880b1

### DIFF
--- a/expected/function/query-expand/multiple-terms.out
+++ b/expected/function/query-expand/multiple-terms.out
@@ -3,23 +3,10 @@ CREATE TABLE synonyms (
   synonyms text[]
 );
 CREATE INDEX synonyms_term_index ON synonyms (term);
-DO LANGUAGE plpgsql $$
-BEGIN
-	PERFORM 1
-		WHERE current_setting('server_version_num')::int >= 120000;
-	IF FOUND THEN
-		INSERT INTO synonyms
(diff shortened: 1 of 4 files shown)


abetomo (approved):

komainu8 (approved):

expected/function/query-expand/multiple-terms.out
```diff
@@ -3,23 +3,10 @@ CREATE TABLE synonyms (
...
-		INSERT INTO synonyms
-			VALUES ('Groonga', ARRAY['Mroonga', 'PGroonga', 'Rroonga']);
-	ELSE
-		INSERT INTO synonyms
-			VALUES ('Groonga', ARRAY['Mroonga', 'PGroonga', 'Rroonga']);
-		INSERT INTO synonyms
-			VALUES ('Groonga', ARRAY['Groonga', 'Senna']);
-	END IF;
-END;
-$$;
```
otegami, line 22: This change is related to the following commit.
- https://github.com/pgroonga/pgroonga/commit/b5bba12cbbb3dbcbb3d5f2c50e434b48a43880b1

PR #640, opened 2024-12-18
test: Remove `*.out` without `*.sql` pairs under `expected/compatibility/`

abetomo:
Related: GitHub GH-637
Since this is a test for compatibility, do not add `*.sql`, only remove it.


kou:
なんでこんなにあったんだ。。。

PR #645, opened 2025-01-17
Remove "pgroonga" scheam

komainu8:
We have already used "pgroonga_" prefix instead.
This modifications are incompatibility.

TODO:
Drop function, operator, and operator class in upgrade sql.
Create function, operator, and operator class in downgrade sql.


komainu8:
I'll split this modification in the other PRs. So, I close this PR.

Issue #647, opened 2025-01-22
Remove "pgroonga" schema


komainu8:
### What is your problem?

We have already used "pgroonga_" prefix instead.
This modifications are incompatibility with PGroonga 3.x or earlier.

We must remove the following schema, operator families, operator classes, operators, and functions.

### Operator families and Operator class

- [x] pgroonga.text_full_text_search_ops FOR TYPE text: #651
- [x] pgroonga.text_array_full_text_search_ops: #654 
- [x] pgroonga.varchar_full_text_search_ops FOR TYPE varchar: #655 
- [x] pgroonga.varchar_ops FOR TYPE varchar: #656
- [x] pgroonga.varchar_array_ops: #657 
- [x] pgroonga.bool_ops FOR TYPE bool: #658
- [x] pgroonga.int2_ops FOR TYPE int2: #659 
- [x] pgroonga.int4_ops FOR TYPE int4: #660 
- [x] pgroonga.int8_ops FOR TYPE int8: #663 
- [x] pgroonga.float4_ops FOR TYPE float4: #664 
- [x] pgroonga.float8_ops FOR TYPE float8: #680 
- [x] pgroonga.timestamp_ops FOR TYPE timestamp: #666 
- [x] pgroonga.timestamptz_ops FOR TYPE timestamptz: #667 
- [x] pgroonga.jsonb_ops FOR TYPE jsonb: #668 
- [x] pgroonga.text_regexp_ops FOR TYPE text: #669 
- [x] pgroonga.varchar_regexp_ops FOR TYPE varchar: #670 
- [x] pgroonga.text_full_text_search_ops_v2 FOR TYPE text: #671 
- [x] pgroonga.text_array_full_text_search_ops_v2: #672 
- [x] pgroonga.text_term_search_ops_v2 FOR TYPE text: #673 
- [x] pgroonga.text_array_term_search_ops_v2 FOR TYPE text[]: #674 
- [x] pgroonga.text_regexp_ops_v2 FOR TYPE text: #675 
- [x] pgroonga.varchar_full_text_search_ops_v2: #676 
- [x] pgroonga.varchar_array_term_search_ops_v2: #678 
- [x] pgroonga.varchar_regexp_ops_v2 FOR TYPE varchar: #679 
- [x] pgroonga.jsonb_ops_v2: #677 

### Operators

none.

### Functions

- [ ] pgroonga.score("row" record): #681 
- [ ] pgroonga.table_name(indexName cstring): #682 
[...]

PR #700, opened 2025-02-05
pgroonga: remove pgroonga.escape(value int8) function

komainu8:
GitHub: GH-647

This is part of the task of remove "pgroonga" schema. It's deprecated since PGroonga 2.
This is incompatible with PGroonga 3.x or earlier.

TODO:
* Test
* Resolve confrict


komainu8:
This PR duplicate to #706.

PR #716, opened 2025-02-19
Optimaize to "match in" from "column %% keyword1 OR column %% keyword2 OR ..."

komainu8:
PostgreSQL 18 optimaize to "column IN (keyword1, keyword2, ...)" from "column %% keyword1 OR column %% keyword2 OR ...".
%% is match operator. So, we handle "column %% keyword1 OR column %% keyword2 OR ..." as "match in" operator.

### DIFF
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -5391,8 +5391,8 @@ static bool
 PGrnSearchIsMatchIn(ScanKey key)
 {
 	return (key->sk_flags & SK_SEARCHARRAY) &&
-		   ((key->sk_strategy == PGrnMatchStrategyNumber)
-			|| (key->sk_strategy == PGrnMatchStrategyV2Number));
+		   ((key->sk_strategy == PGrnMatchStrategyNumber) ||
+			(key->sk_strategy == PGrnMatchStrategyV2Number));
 }
 
 static void
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -6129,10 +6129,11 @@ PGrnSearchBuildCondition(Relation index, ScanKey key, PGrnSearchData *data)
 	}
 	if (PGrnSearchIsMatchIn(key))
 	{
-		if (key->sk_strategy != PGrnMatchInStrategyV2Number)
-		{
-			key->sk_strategy = PGrnMatchInStrategyV2Number;
-		}
+		// PostgreSQL 18 optimaize to "column IN (keyword1, keyword2, ...)" from
+		// "column %% keyword1 OR column %% keyword2 OR ...".
+		// %% is match operator. So, we handle "column %% keyword1 OR column %%
+		// keyword2 OR ..." as "match in" operator.
+		key->sk_strategy = PGrnMatchInStrategyV2Number;
 	}
 
 	if (key->sk_strategy == PGrnMatchFTSConditionStrategyV2Number ||
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -53,7 +53,6 @@
 #include <nodes/nodeFuncs.h>
 #include <optimizer/optimizer.h>
 #include <pgstat.h>
-#include <stdbool.h>
 #include <storage/bufmgr.h>
 #include <storage/ipc.h>
 #include <storage/latch.h>
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -53,6 +53,7 @@
 #include <nodes/nodeFuncs.h>
 #include <optimizer/optimizer.h>
 #include <pgstat.h>
+#include <stdbool.h>
 #include <storage/bufmgr.h>
 #include <storage/ipc.h>
 #include <storage/latch.h>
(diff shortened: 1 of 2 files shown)


komainu8:
Unit tests fail on PostgreSQL 18 snapshot.
However, this failure don't related this modification.
Because the failed test occurs when `pgroonga_command()` execution. This modification doesn't related `pgroonga_command()`.

src/pgroonga.c
```diff
@@ -6119,6 +6127,14 @@ PGrnSearchBuildCondition(Relation index, ScanKey key, PGrnSearchData *data)
 		PGrnSearchBuildConditionIn(data, key, targetColumn, attribute);
 		return;
 	}
+	if (PGrnSearchIsMatchIn(key))
```
otegami, line 6130: Is it possible that we expand the existed logic of `PGrnSearchIsInCondition` instead of defining `PGrnSearchIsMatchIn`?

komainu8: I don't unify this logic.
Because `IN` and `IN + Match` is different.

otegami (approved):
LGTM 👍🏾

src/pgroonga.c
otegami: Thank you so much. I understood.

src/pgroonga.c
```diff
@@ -6119,6 +6127,14 @@ PGrnSearchBuildCondition(Relation index, ScanKey key, PGrnSearchData *data)
 		PGrnSearchBuildConditionIn(data, key, targetColumn, attribute);
 		return;
 	}
+	if (PGrnSearchIsMatchInCondition(key))
+	{
+		// PostgreSQL 18 optimaize to "column IN (keyword1, keyword2, ...)" from
+		// "column %% keyword1 OR column %% keyword2 OR ...".
+		// %% is match operator. So, we handle "column %% keyword1 OR column %%
+		// keyword2 OR ..." as "match in" operator.
+		key->sk_strategy = PGrnMatchInStrategyV2Number;
```
abetomo, line 6136: How about adding a PGrnSearchBuildMatchConditionIn() like PGrnSearchIsInCondition() and PGrnSearchBuildConditionIn()?

src/pgroonga.c
```diff
@@ -6119,6 +6127,14 @@ PGrnSearchBuildCondition(Relation index, ScanKey key, PGrnSearchData *data)
 		PGrnSearchBuildConditionIn(data, key, targetColumn, attribute);
 		return;
 	}
+	if (PGrnSearchIsMatchInCondition(key))
+	{
+		// PostgreSQL 18 optimaize to "column IN (keyword1, keyword2, ...)" from
+		// "column %% keyword1 OR column %% keyword2 OR ...".
+		// %% is match operator. So, we handle "column %% keyword1 OR column %%
+		// keyword2 OR ..." as "match in" operator.
```
kou, line 6135: Is it really `IN`? `column &@ ANY (keyword1, keyword2, ...)`?

BTW, why did you use deprecated `%%` not `&@`?

otegami:
It looks nice to me except for this https://github.com/pgroonga/pgroonga/pull/716#discussion_r1970850306 comment.
If it's still in progress, I'm so sorry mm.

komainu8:
Thank you for your comment.
I apply https://github.com/pgroonga/pgroonga/pull/716#discussion_r1970850306 comment yet.
If it's still in progress.

src/pgroonga.c
komainu8: I added a `PGrnSearchBuildConditionMatchIn()`.

otegami: What do you think of renaming this function to `PGrnSearchSetMatchInStrategy()` ?
Regarding `PGrnSearchBuildConditionMatchIn()`, the function name suggests that it builds a condition.
But it only modifies the parameters (specifically, the sk_strategy). Compared to other functions in the `BuildCondition` family, its behavior is inconsistent.

komainu8: Thank you for your comment.
I've fixed it.

otegami (approved):
LGTM if the failed CI cases are as expected.

src/pgroonga.c
```diff
@@ -6095,6 +6103,17 @@ PGrnSearchBuildConditionBinaryOperation(PGrnSearchData *data,
...
 
+static void
+PGrnSearchSetMatchInStrategy(ScanKey key)
+{
+	// PostgreSQL 18 optimaize to "column &@ ANY (keyword1, keyword2, ...)"
+	// from "column &@ keyword1 OR column &@ keyword2 OR ...".
+	// &@ is match operator. So, we should handle
+	// "column &@ keyword1 OR column &@ keyword2 OR ..." as "match in"
+	// operator.
+	key->sk_strategy = PGrnMatchInStrategyV2Number;
```
kou, line 6114: In general, we should not rewrite the given `ScanKey`... We should use it as a read-only data because it may be used by caller later.

Can we avoid rewriting the given `ScanKey`?

komainu8:
The following pattern return wrong result as below in PostgreSQL 18.

```sql
CREATE TABLE tags (
  id int,
  name text
);

INSERT INTO tags VALUES (1, 'PostgreSQL');
INSERT INTO tags VALUES (2, 'Groonga');
INSERT INTO tags VALUES (3, 'groonga');
INSERT INTO tags VALUES (4, 'PGroonga');

CREATE INDEX pgrn_index ON tags
  USING pgroonga (name pgroonga_text_term_search_ops_v2);

SET enable_seqscan = off;
SET enable_indexscan = on;
SET enable_bitmapscan = off;

SELECT name
  FROM tags
 WHERE name &= 'groonga' OR name &= 'PGroonga'
 ORDER BY id;
-- name 
-- ------
-- (0 行)
```

komainu8:
I will fix the failing tests.

PR #718, opened 2025-02-21
Add pgroonga.enable_custom_scan parameter

abetomo:
Preparing to implement a custom scan for PGroonga.
This parameter is used to enable or disable custom scan.

The description will be added in detail when implementation is complete.


makefiles/pgroonga-sources.mk
```diff
@@ -5,6 +5,7 @@ SRCS =						\
 	src/pgrn-convert.c			\
 	src/pgrn-create.c			\
 	src/pgrn-ctid.c				\
+	src/pgrn-customscan.c			\
```
kou, line 8: `pgrn-custom-scan.c`がいいかな。

PR #720, opened 2025-02-21
ci: Work around permission error

abetomo:
Permission error.

```
mkdir: cannot create directory ‘/host-rw/logs’: Permission denied
```

To work around this, create `logs` directory on the host side with 777.

PR #721, opened 2025-02-22
ci: Add `shift=true` option to `incus config device add`

abetomo:
Permission error:

```
mkdir: cannot create directory ‘/host-rw/logs’: Permission denied
```

Adding the `shift=true` option solves this.

shift:
Sets up a shifting overlay to translate the source UID/GID to match the instance (only for containers)
https://linuxcontainers.org/incus/docs/main/reference/devices_disk/

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -178,7 +178,7 @@ jobs:
           set -x
           sudo incus admin init --auto
           sudo incus launch --quiet ${{ env.TEST_INCUS_IMAGE }} target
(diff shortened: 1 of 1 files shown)


abetomo:
Related: #720

komainu8 (approved):

otegami (approved):
LGTM 👍🏾 
📝  `shift` option doesn't work if we launch virtual machine using `--vm` option.

kou:
Confirmed: https://github.com/pgroonga/pgroonga/actions/runs/13490019971/job/37826762833?pr=721#step:14:3493

PR #722, opened 2025-03-04
Add skeletons to implement custom scan

abetomo:
Prepare to implement custom scan.
https://www.postgresql.org/docs/current/custom-scan.html

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -1,3 +1,10 @@
+#include <postgres.h>
+
+#include <nodes/extensible.h>
+#include <optimizer/pathnode.h>
+#include <optimizer/paths.h>
+#include <optimizer/restrictinfo.h>
+
 #include "pgrn-custom-scan.h"
 #include "pgroonga.h"
 
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -171,3 +171,8 @@ PGrnInitializeCustomScan(void)
 
 	RegisterCustomScanMethods(&PGrnScanMethods);
 }
+
+void
+PGrnFinalizeCustomScan(void)
+{
+}
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -175,4 +175,9 @@ PGrnInitializeCustomScan(void)
 void
 PGrnFinalizeCustomScan(void)
 {
+	set_rel_pathlist_hook = PreviousSetRelPathlistHook;
+	// todo
+	// Disable registered functions.
+	// See also
+	// https://github.com/pgroonga/pgroonga/pull/722#discussion_r2011284556
 }
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -1,6 +1,58 @@
 #include "pgroonga.h"
+#include "pgrn-custom-scan.h"
+
+typedef struct PGrnScanState
+{
+	CustomScanState customScanState; /* must be first field */
+} PGrnScanState;
 
 static bool PGrnCustomScanEnabled = false;
+static set_rel_pathlist_hook_type PreviousSetRelPathlistHook = NULL;
+
+static void PGrnSetRelPathlistHook(PlannerInfo *root,
+								   RelOptInfo *rel,
+								   Index rti,
+								   RangeTblEntry *rte);
+static Plan *PGrnPlanCustomPath(PlannerInfo *root,
+								RelOptInfo *rel,
+								struct CustomPath *best_path,
+								List *tlist,
+								List *clauses,
+								List *custom_plans);
+static List *PGrnReparameterizeCustomPathByChild(PlannerInfo *root,
+												 List *custom_private,
+												 RelOptInfo *child_rel);
+static Node *PGrnCreateCustomScanState(CustomScan *cscan);
+static void
+PGrnScanBeginCustomScan(CustomScanState *node, EState *estate, int eflags);
+static TupleTableSlot *PGrnScanExecCustomScan(CustomScanState *node);
+static void PGrnScanEndCustomScan(CustomScanState *node);
+static void PGrnScanReScanCustomScan(CustomScanState *node);
+static void PGrnScanExplainCustomScan(CustomScanState *node,
+									  List *ancestors,
+									  ExplainState *es);
+
+const struct CustomPathMethods PGrnScanPathMethods = {
+	.CustomName = "PGroongaScan",
+	.PlanCustomPath = PGrnPlanCustomPath,
+	.ReparameterizeCustomPathByChild = PGrnReparameterizeCustomPathByChild,
+};
+
+const struct CustomScanMethods PGrnScanScanMethods = {
+	.CustomName = "PGroongaScan",
+	.CreateCustomScanState = PGrnCreateCustomScanState,
+};
+
+const struct CustomExecMethods PGrnScanExecuteMethods = {
+	.CustomName = "PGroongaScan",
+
+	.BeginCustomScan = PGrnScanBeginCustomScan,
+	.ExecCustomScan = PGrnScanExecCustomScan,
+	.EndCustomScan = PGrnScanEndCustomScan,
+	.ReScanCustomScan = PGrnScanReScanCustomScan,
+
+	.ExplainCustomScan = PGrnScanExplainCustomScan,
+};
 
 bool
 PGrnCustomScanGetEnabled(void)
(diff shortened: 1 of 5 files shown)


src/pgrn-custom-scan.h
```diff
@@ -1,3 +1,12 @@
+#include <postgres.h>
+
+#include <nodes/extensible.h>
+#include <optimizer/pathnode.h>
+#include <optimizer/paths.h>
+#include <optimizer/restrictinfo.h>
```
kou, line 6: Can we move this to `pgrn-custom-scan.c`?

src/pgrn-custom-scan.c
```diff
@@ -1,6 +1,58 @@
...
+									  List *ancestors,
+									  ExplainState *es);
+
+const struct CustomPathMethods PGrnScanPathMethods = {
+	.CustomName = "PGroongaScan",
+	.PlanCustomPath = PGrnPlanCustomPath,
+	.ReparameterizeCustomPathByChild = PGrnReparameterizeCustomPathByChild,
+};
+
+const struct CustomScanMethods PGrnScanScanMethods = {
```
kou, line 41: `PGrnScanMethods`とか？

src/pgrn-custom-scan.c
```diff
@@ -1,6 +1,58 @@
...
+	.CustomName = "PGroongaScan",
+	.CreateCustomScanState = PGrnCreateCustomScanState,
+};
+
+const struct CustomExecMethods PGrnScanExecuteMethods = {
```
kou, line 46: `PGrnExecuteMethods`とか？

src/pgrn-custom-scan.c
```diff
@@ -1,6 +1,58 @@
...
+	.CustomName = "PGroongaScan",
+
+	.BeginCustomScan = PGrnScanBeginCustomScan,
```
kou, line 49: `PGrnBeginCustomScan`とか？

src/pgrn-custom-scan.c
```diff
@@ -19,3 +71,100 @@ PGrnCustomScanDisable(void)
...
+PGrnPlanCustomPath(PlannerInfo *root,
+				   RelOptInfo *rel,
+				   struct CustomPath *best_path,
+				   List *tlist,
+				   List *clauses,
+				   List *custom_plans)
+{
+	CustomScan *cscan = makeNode(CustomScan);
+	cscan->methods = &PGrnScanScanMethods;
+	return (Plan *) cscan;
```
kou, line 111: Is it correct? `&(cscan->scan.plan)`?

src/pgrn-custom-scan.c
```diff
@@ -19,3 +71,100 @@ PGrnCustomScanDisable(void)
...
+}
+
+static List *
+PGrnReparameterizeCustomPathByChild(PlannerInfo *root,
+									List *custom_private,
+									RelOptInfo *child_rel)
+{
+	return NULL;
```
kou, line 119:
```suggestion
	return NIL;
```

src/pgrn-custom-scan.c
```diff
@@ -19,3 +71,100 @@ PGrnCustomScanDisable(void)
...
+}
+
+static Node *
+PGrnCreateCustomScanState(CustomScan *cscan)
+{
+	PGrnScanState *pgrnScanState =
```
kou, line 125:
```suggestion
	PGrnScanState *state =
```

src/pgrn-custom-scan.c
```diff
@@ -1,6 +1,58 @@
+#include "pgrn-custom-scan.h"
 #include "pgroonga.h"
 
+typedef struct PGrnScanState
+{
+	CustomScanState customScanState; /* must be first field */
```
kou, line 6: How about renaming this to `parent`?

```suggestion
	CustomScanState parent; /* must be first field */
```

src/pgrn-custom-scan.c
```diff
@@ -19,3 +71,100 @@ PGrnCustomScanDisable(void)
...
+}
+
+static Node *
+PGrnCreateCustomScanState(CustomScan *cscan)
+{
+	PGrnScanState *pgrnScanState =
+		(PGrnScanState *) newNode(sizeof(PGrnScanState), T_CustomScanState);
+
+	CustomScanState *cscanstate = &pgrnScanState->customScanState;
+	cscanstate->methods = &PGrnScanExecuteMethods;
```
kou, line 129:
```suggestion
	state->parent.methods = &PGrnScanExecuteMethods;
```

src/pgrn-custom-scan.c
```diff
@@ -19,3 +71,100 @@ PGrnCustomScanDisable(void)
...
+
+	return (Node *) cscanstate;
```
kou, line 131:
```suggestion
	return (Node *) state;
```

src/pgroonga.c
```diff
@@ -738,6 +739,8 @@ _PG_init(void)
 
 		PGrnInitializeOptions();
 
+		PGrnInitializeCustomScan();
```
kou, line 742: Can we finalize in `PGrnBeforeShmemExit()`?

src/pgrn-custom-scan.c
abetomo: I have defined `PGrnScanState` to add a PGroonga specific field to `CustomScanState`.
(There are no specific fields yet, but we will add them in the implementation.)
So I think `customScanState` (or `state` or something) would be better, since it should be the current state, not the parent, what do you think?

kou: In general, we can use cast to access parent struct in C:

```c
PGrnScanState *state = palloc(sizeof(PGrnScanState));
CustomScanState *base_state = (CustomScanState *)state;
base_state->XXX = ...;
```

We don't need to use the member to use `CustomScanState`:

```c
PGrnScanState *state = palloc(sizeof(PGrnScanState));
state->customScanState.XXX = ...;
```

abetomo: I didn't know that this type of structure usage is called parent and child.
So I was misunderstanding it.

The name `parent` made sense to me.

kou: It's a generic `struct` usage to use object oriented programming in C.
"parent" is for "parent class".

src/pgrn-custom-scan.c
```diff
@@ -1,6 +1,64 @@
...
+#include <optimizer/paths.h>
+#include <optimizer/restrictinfo.h>
+
+#include "pgrn-custom-scan.h"
 #include "pgroonga.h"
 
+typedef struct PGrnScanState
+{
+	CustomScanState parent; /* must be first field */
+} PGrnScanState;
```
otegami, line 14: 📝 Wrapping the `CustomScanState` for adding the PGroonga properties later.

src/pgrn-custom-scan.c
```diff
@@ -19,3 +77,102 @@ PGrnCustomScanDisable(void)
...
+
+static Node *
+PGrnCreateCustomScanState(CustomScan *cscan)
+{
+	PGrnScanState *state =
+		(PGrnScanState *) newNode(sizeof(PGrnScanState), T_CustomScanState);
+
+	state->parent.methods = &PGrnExecuteMethods;
+
+	return (Node *) &(state->parent);
```
kou, line 136:
```suggestion
	return (Node *) &state;
```

src/pgrn-custom-scan.c
```diff
@@ -19,3 +77,102 @@ PGrnCustomScanDisable(void)
...
+	PreviousSetRelPathlistHook = set_rel_pathlist_hook;
+	set_rel_pathlist_hook = PGrnSetRelPathlistHook;
+
+	RegisterCustomScanMethods(&PGrnScanMethods);
+}
+
+void
+PGrnFinalizeCustomScan(void)
+{
+}
```
kou, line 178: Do we need `set_rel_pathlist_hook = PreviousSetRelPathlistHook` here?

otegami (commented):
📝memo

I understood we don't need the following callbacks because we will consider if the `CUSTOMPATH_SUPPORT_MARK_RESTORE` and `CUSTOMPATH_SUPPORT_MARK_RESTORE` flag are needed.
```c
void (*MarkPosCustomScan) (CustomScanState *node);
void (*RestrPosCustomScan) (CustomScanState *node);
```

ref: https://www.postgresql.org/docs/current/custom-scan-execution.html

At this moment, we don't consider the parallel operations.
So we don't need the following functions now.

```c
void (*InitializeDSMCustomScan) (CustomScanState *node,
                                 ParallelContext *pcxt,
                                 void *coordinate);
void (*ReInitializeDSMCustomScan) (CustomScanState *node,
                                   ParallelContext *pcxt,
                                   void *coordinate);
void (*InitializeWorkerCustomScan) (CustomScanState *node,
                                    shm_toc *toc,
                                    void *coordinate);
void (*ShutdownCustomScan) (CustomScanState *node);
```

ref: https://www.postgresql.org/docs/current/custom-scan-execution.html

src/pgrn-custom-scan.c
abetomo: I could not find any documentation if `set_rel_pathlist_hook = PreviousSetRelPathlistHook` is needed or not.

However, there is no UnregisterCustomScanMethods() pair to RegisterCustomScanMethods(), and registered CustomScanMethods cannot be reset.
https://github.com/postgres/postgres/blob/cb36f8ec2124b0c495da057a2f55735fe858be0e/src/include/nodes/extensible.h#L160
So I have decided that I don't need to reset the `set_rel_pathlist_hook` too.

kou: Hmm. We had some crash bugs that are caused by running some registered functions are used after PGroonga is finalized.

Can we ensure disabling all registered functions (exits immediately?) after PGroonga is finalized?

abetomo: I got it.
I will implement function disabling in a new PR.

PR #726, opened 2025-04-02
ci: windows: Add `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` option to CMake

abetomo:
CMake in CI environment fails with 4.0.0.

```
CMake Error at D:/a/pgroonga/build/_deps/zstd-src/build/cmake/CMakeLists.txt:10 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

-- Configuring incomplete, errors occurred!
```

We plan to migrate to Meson in the future.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -128,6 +128,7 @@ jobs:
             -S . ^
             -DCMAKE_BUILD_TYPE=RelWithDebInfo ^
             -DCMAKE_INSTALL_PREFIX="..\pgsql" ^
+            -DCMAKE_POLICY_VERSION_MINIMUM=3.5 ^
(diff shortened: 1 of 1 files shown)


kou:
It seems that we can fix by updating bundled Zstandard: https://github.com/groonga/groonga/blob/f2d3441a94af5ed557a273ddc62104daaea38ecd/CMakeLists.txt#L642-L644

PR #734, opened 2025-04-25
ci: Fix incus device add error

abetomo:
With the latest GitHub Actions image, we get the following error:

```
Error: Failed to start device "host-rw": Required idmapping abilities not available
```

Same issue as this one:
https://discuss.linuxcontainers.org/t/error-failed-to-setup-device-mount-idmapping-abilities-are-required-but-arent-supported-on-system/19976

If the `shift=true` option is removed, this error will not occur.
However, when that option is removed, the GH-721 issue reoccurs.

That permission error could be avoided by creating the directory with mode=777.
So, remove the `shift=true` option and create the directory with mode=777.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -178,7 +178,8 @@ jobs:
           set -x
           sudo incus admin init --auto
           sudo incus launch --quiet ${{ env.TEST_INCUS_IMAGE }} target
-          sudo incus config device add target host-rw disk shift=true source=$PWD path=/host-rw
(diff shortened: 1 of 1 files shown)


abetomo:
Error if image version is 20250420.1.0 or later.
https://github.com/pgroonga/pgroonga/actions/runs/14654953991/job/41128296705#step:1:9

When I ran CI with the test always failing, the logs were uploaded to the artifact.
https://github.com/pgroonga/pgroonga/actions/runs/14654953991

otegami (approved):
LGTM 👍🏾

PR #738, opened 2025-05-09
ci package almalinux8/9: add/update Apache Arrow repository

otegami:
## Problem

The following error in build step is raised because Groonga APT repository doesn't serve arrow-devel-x.x.x package.

```
69.33 Error:
69.33  Problem: cannot install the best candidate for the job
69.33   - nothing provides arrow-devel >= 20.0.0 needed by groonga-devel-15.0.9-1.el8.x86_64 from groonga-almalinux
69.33 (try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

## Solution

We added/updated the official Apache Arrow repository.


komainu8:
Could you modify `packages/yum/test.sh` too?

otegami:
> Could you modify packages/yum/test.sh too?

fix: [f966ef0](https://github.com/pgroonga/pgroonga/pull/738/commits/f966ef042c1318ec6151f72db92d2526137d0aa4) Thanks, I updated it.

otegami:
Thank you for reviewing. I've just addressed all the review comments.

komainu8 (approved):

PR #740, opened 2025-05-23
custom-scan: Add PGrnChooseIndex()

abetomo:

Custom scan needs to implement logic to choose which indexes to use.
Because our first goal is to be able to return records, we choose the first PGroonga index we find in the for loop in a temporary implementation.

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -6,8 +6,10 @@
 #include <optimizer/restrictinfo.h>
 #include <utils/lsyscache.h>
 
-#include "pgrn-custom-scan.h"
 #include "pgroonga.h"
+
+#include "pgrn-custom-scan.h"
+#include "pgrn-groonga.h"
 
 typedef struct PGrnScanState
 {
@@ -147,9 +149,43 @@ PGrnCreateCustomScanState(CustomScan *cscan)
 	return (Node *) &(state->parent);
 }
 
-static void
-PGrnBeginCustomScan(CustomScanState *cscanstate, EState *estate, int eflags)
+static Relation
+PGrnChooseIndex(Relation table_rel, int errorLevel)
 {
+	// todo: Support pgroonga_condition() index specification.
+	// todo: Implementation of the logic for choosing which index to use.
+	ListCell *lc;
+	List *indexes;
+
+	if (!table_rel)
+		return NULL;
+
+	indexes = RelationGetIndexList(table_rel);
+	foreach (lc, indexes)
+	{
+		Oid indexId = lfirst_oid(lc);
+
+		Relation index = RelationIdGetRelation(indexId);
+		if (!PGrnIndexIsPGroonga(index))
+		{
+			RelationClose(index);
+			continue;
+		}
+		return index;
+	}
+	return NULL;
+}
+
+static void
+PGrnBeginCustomScan(CustomScanState *node, EState *estate, int eflags)
+{
+	PGrnScanState *state = (PGrnScanState *) node;
+	CustomScanState *customScanState = (CustomScanState *) state;
+	Relation index =
+		PGrnChooseIndex(customScanState->ss.ss_currentRelation, ERROR);
+	grn_obj *sourcesTable = PGrnLookupSourcesTable(index, ERROR);
+	elog(LOG, "DEBUG: custom-scan: %s", PGrnInspect(sourcesTable));
+	RelationClose(index);
 }
 
 static TupleTableSlot *


src/pgrn-custom-scan.c
```diff
@@ -147,9 +149,43 @@ PGrnCreateCustomScanState(CustomScan *cscan)
...
-PGrnBeginCustomScan(CustomScanState *cscanstate, EState *estate, int eflags)
+PGrnBeginCustomScan(CustomScanState *node, EState *estate, int eflags)
 {
+	PGrnScanState *state = (PGrnScanState *) node;
+	CustomScanState *customScanState = (CustomScanState *) state;
+	Relation index =
+		PGrnChooseIndex(customScanState->ss.ss_currentRelation, ERROR);
+	grn_obj *sourcesTable = PGrnLookupSourcesTable(index, ERROR);
+	elog(LOG, "DEBUG: custom-scan: %s", PGrnInspect(sourcesTable));
+	RelationClose(index);
```
otegami, line 188: I’m not sure whether we need to close the `index` in the `EndCustomScan` step.
According to the PostgreSQL documentation, this index value isn’t stored in the `CustomScanState *node`,
so we don’t need to worry about closing it, do we?
https://www.postgresql.org/docs/current/custom-scan-execution.html#CUSTOM-SCAN-EXECUTION-CALLBACKS

abetomo: This `index` variable does not need to be closed in EndCustomScan().
This is because it is only used in PGrnBeginCustomScan() and closed here.

EndCustomScan() closes the resource set to `CustomScanState` (`PGrnScanState` in this code) as you said.
There is currently no value set for `PGrnScanState`, so EndCustomScan() does not close anything.

src/pgrn-custom-scan.c
```diff
@@ -147,9 +149,43 @@ PGrnCreateCustomScanState(CustomScan *cscan)
 	return (Node *) &(state->parent);
 }
 
+static Relation
+PGrnChooseIndex(Relation table_rel, int errorLevel)
+{
+	// todo: Support pgroonga_condition() index specification.
+	// todo: Implementation of the logic for choosing which index to use.
+	ListCell *lc;
```
kou, line 157:
```suggestion
	ListCell *cell;
```

src/pgrn-custom-scan.c
```diff
@@ -147,9 +149,43 @@ PGrnCreateCustomScanState(CustomScan *cscan)
 	return (Node *) &(state->parent);
 }
 
+static Relation
+PGrnChooseIndex(Relation table_rel, int errorLevel)
```
kou, line 153:
```suggestion
PGrnChooseIndex(Relation table, int errorLevel)
```

src/pgrn-custom-scan.c
```diff
@@ -147,9 +149,43 @@ PGrnCreateCustomScanState(CustomScan *cscan)
...
+	}
+	return NULL;
+}
+
 static void
-PGrnBeginCustomScan(CustomScanState *cscanstate, EState *estate, int eflags)
+PGrnBeginCustomScan(CustomScanState *node, EState *estate, int eflags)
 {
+	PGrnScanState *state = (PGrnScanState *) node;
+	CustomScanState *customScanState = (CustomScanState *) state;
```
kou, line 183:
```suggestion
PGrnBeginCustomScan(CustomScanState *customScanState, EState *estate, int eflags)
{
	PGrnScanState *state = (PGrnScanState *) customScanState;
```

src/pgrn-custom-scan.c
kou, line 153: It seems that `int errorLevel` isn't used.

otegami (approved):

src/pgrn-custom-scan.c
```diff
@@ -147,9 +149,42 @@ PGrnCreateCustomScanState(CustomScan *cscan)
...
+	return NULL;
+}
+
 static void
-PGrnBeginCustomScan(CustomScanState *cscanstate, EState *estate, int eflags)
+PGrnBeginCustomScan(CustomScanState *customScanState,
+					EState *estate,
+					int eflags)
 {
+	Relation index = PGrnChooseIndex(customScanState->ss.ss_currentRelation);
```
komainu8, line 184:
```suggestion
	Relation index = PGrnChooseIndex(customScanState->ss.ss_currentRelation);
	if (!index)
	  return;
```

Probably, if `index` is `NULL`, PGroonga crash in `PGrnLookupSourcesTable()`

src/pgrn-custom-scan.c
```diff
@@ -147,9 +149,42 @@ PGrnCreateCustomScanState(CustomScan *cscan)
...
+static Relation
+PGrnChooseIndex(Relation table)
+{
+	// todo: Support pgroonga_condition() index specification.
+	// todo: Implementation of the logic for choosing which index to use.
+	ListCell *cell;
+	List *indexes;
+
+	if (!table)
+		return NULL;
```
komainu8, line 161: How `return NIL` and `return NULL` are different?
Because `PGrnReparameterizeCustomPathByChild()` returns `NIL`.

abetomo: `NIL` means an empty `List` type.
So, `PGrnReparameterizeCustomPathByChild()`, whose return value is of `List` type, returns `NIL`.

PR #742, opened 2025-05-26
custom-scan: Fetch and return data from Groonga resources

abetomo:

The first goal is to get the results of the following query as expected.
```sql
/* sql/custom-scan/debug.sql */
 SELECT content
   FROM memos
  WHERE content &@~ 'PGroonga OR Groonga';
```

This commit does not consider the conditions of the `WHERE` clause and fetches all data from Groonga's sources table.
The goal of the next PR is to filter by the conditions of the `WHERE` clause.

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -218,7 +218,7 @@ PGrnExecCustomScan(CustomScanState *customScanState)
 		return NULL;
 
 	id = grn_table_cursor_next(ctx, state->tableCursor);
-	if (!id)
+	if (id == GRN_ID_NIL)
 		return NULL;
 
 	{
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -260,7 +260,7 @@ PGrnEndCustomScan(CustomScanState *customScanState)
 		grn_table_cursor_close(ctx, state->tableCursor);
 		state->tableCursor = NULL;
 	}
-	nTargetColumns = GRN_BULK_VSIZE(&(state->columns)) / sizeof(grn_obj *);
+	nTargetColumns = GRN_PTR_VECTOR_SIZE(&(state->columns));
 	for (unsigned int i = 0; i < nTargetColumns; i++)
 	{
 		grn_obj *column = GRN_PTR_VALUE_AT(&(state->columns), i);
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -148,6 +148,9 @@ PGrnCreateCustomScanState(CustomScan *cscan)
 
 	state->parent.methods = &PGrnExecuteMethods;
 
+	state->tableCursor = NULL;
+	GRN_PTR_INIT(&(state->columns), GRN_OBJ_VECTOR, GRN_ID_NIL);
+
 	return (Node *) &(state->parent);
 }
 
@@ -194,7 +197,6 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
 	sourcesTable = PGrnLookupSourcesTable(index, ERROR);
 	RelationClose(index);
 
-	GRN_PTR_INIT(&(state->columns), GRN_OBJ_VECTOR, GRN_ID_NIL);
 	for (unsigned int i = 0; i < tupdesc->natts; i++)
 	{
 		Form_pg_attribute attr = TupleDescAttr(tupdesc, i);
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -231,6 +231,7 @@ PGrnExecCustomScan(CustomScanState *customScanState)
 		{
 			Form_pg_attribute attr = &(slot->tts_tupleDescriptor->attrs[i]);
 			grn_obj *column = GRN_PTR_VALUE_AT(&(state->columns), i);
+			GRN_BULK_REWIND(&value);
 			grn_obj_get_value(ctx, column, id, &value);
 			slot->tts_values[i] = PGrnConvertToDatum(&value, attr->atttypid);
 			slot->tts_isnull[i] = false;
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -235,6 +235,10 @@ PGrnExecCustomScan(CustomScanState *customScanState)
 			grn_obj_get_value(ctx, column, id, &(state->columnValue));
 			slot->tts_values[i] =
 				PGrnConvertToDatum(&(state->columnValue), attr->atttypid);
+			// todo
+			// If there are nullable columns, do not custom scan.
+			// See also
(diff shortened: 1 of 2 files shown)


src/pgrn-custom-scan.c
```diff
@@ -181,19 +183,59 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
...
 	Relation index = PGrnChooseIndex(customScanState->ss.ss_currentRelation);
+	grn_obj *sourcesTable;
+
 	if (!index)
 		return;
 	sourcesTable = PGrnLookupSourcesTable(index, ERROR);
-	elog(LOG, "DEBUG: custom-scan: %s", PGrnInspect(sourcesTable));
 	RelationClose(index);
+
+	GRN_PTR_INIT(&(state->columns), GRN_OBJ_VECTOR, GRN_ID_NIL);
```
kou, line 197: We should do this in `PGrnCreateCustomScanState()` to ensure `state->columns` is valid in `PGrnEndCustomScan()`.

src/pgrn-custom-scan.c
```diff
@@ -181,19 +183,59 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
...
 {
-	return NULL;
+	PGrnScanState *state = (PGrnScanState *) customScanState;
+	grn_id id;
+
+	if (!state->tableCursor)
+		return NULL;
+
+	id = grn_table_cursor_next(ctx, state->tableCursor);
+	if (!id)
```
kou, line 219:
```suggestion
	if (id != GRN_ID_NIL)
```

src/pgrn-custom-scan.c
```diff
@@ -181,19 +183,59 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
...
+		return NULL;
+
+	{
+		TupleTableSlot *slot = customScanState->ss.ps.ps_ResultTupleSlot;
+		grn_obj value;
+
+		ExecClearTuple(slot);
+		GRN_VOID_INIT(&value);
```
kou, line 227: Could you move this to `PGrnScanState::column_value` or something and reuse this to avoid allocations as much as possible?

src/pgrn-custom-scan.c
```diff
@@ -181,19 +183,59 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
...
+		for (unsigned int i = 0; i < slot->tts_tupleDescriptor->natts; i++)
+		{
+			Form_pg_attribute attr = &(slot->tts_tupleDescriptor->attrs[i]);
+			grn_obj *column = GRN_PTR_VALUE_AT(&(state->columns), i);
+			grn_obj_get_value(ctx, column, id, &value);
```
kou, line 232:
```suggestion
			GRN_BULK_REWIND(&value);
			grn_obj_get_value(ctx, column, id, &value);
```

src/pgrn-custom-scan.c
```diff
@@ -203,8 +245,21 @@ PGrnExplainCustomScan(CustomScanState *node, List *ancestors, ExplainState *es)
...
 static void
-PGrnEndCustomScan(CustomScanState *node)
+PGrnEndCustomScan(CustomScanState *customScanState)
 {
+	PGrnScanState *state = (PGrnScanState *) customScanState;
+	unsigned int nTargetColumns;
+
+	ExecClearTuple(customScanState->ss.ps.ps_ResultTupleSlot);
+
+	grn_table_cursor_close(ctx, state->tableCursor);
```
kou, line 255: Do we need `if (!state->tableCursor)`?

src/pgrn-custom-scan.c
```diff
@@ -181,19 +183,59 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
...
+
+		ExecClearTuple(slot);
+		GRN_VOID_INIT(&value);
+		for (unsigned int i = 0; i < slot->tts_tupleDescriptor->natts; i++)
+		{
+			Form_pg_attribute attr = &(slot->tts_tupleDescriptor->attrs[i]);
+			grn_obj *column = GRN_PTR_VALUE_AT(&(state->columns), i);
+			grn_obj_get_value(ctx, column, id, &value);
+			slot->tts_values[i] = PGrnConvertToDatum(&value, attr->atttypid);
+			slot->tts_isnull[i] = false;
```
kou, line 234: TODO: We need to reject our custom scan if there is at least one nullable column like we did for index only scan: https://github.com/pgroonga/pgroonga/blob/cac1dbb0fc4aa515595760ad2a0aa3e060594cdd/src/pgroonga.c#L8519

src/pgrn-custom-scan.c
```diff
@@ -181,19 +183,59 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
...
+		}
+		GRN_OBJ_FIN(ctx, &value);
```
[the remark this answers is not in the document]
abetomo: Yes, it is right.
Including the code for handling the conditions would be a large patch, so I would implement it in the next PR.

src/pgrn-custom-scan.c
```diff
@@ -181,19 +183,59 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
...
-	grn_obj *sourcesTable;
+	PGrnScanState *state = (PGrnScanState *) customScanState;
+	TupleDesc tupdesc =
+		RelationGetDescr(customScanState->ss.ss_currentRelation);
 	Relation index = PGrnChooseIndex(customScanState->ss.ss_currentRelation);
+	grn_obj *sourcesTable;
+
 	if (!index)
 		return;
 	sourcesTable = PGrnLookupSourcesTable(index, ERROR);
```
komainu8, line 194:
```suggestion
	sourcesTable = PGrnLookupSourcesTable(index, ERROR);
	if (!sourceTable)
	  return;
```

kou: This is needless because we use `ERROR` for `PGrnLookupSourcesTable()`.

src/pgrn-custom-scan.c
```diff
@@ -203,8 +252,26 @@ PGrnExplainCustomScan(CustomScanState *node, List *ancestors, ExplainState *es)
...
+		state->tableCursor = NULL;
+	}
+	nTargetColumns = GRN_PTR_VECTOR_SIZE(&(state->columns));
+	for (unsigned int i = 0; i < nTargetColumns; i++)
+	{
+		grn_obj *column = GRN_PTR_VALUE_AT(&(state->columns), i);
+		grn_obj_unlink(ctx, column);
+	}
+	GRN_OBJ_FIN(ctx, &(state->columns));
+	GRN_OBJ_FIN(ctx, &(state->columnValue));
```
komainu8, line 274: It would be better to free in order as below.
Because we allocated `state->columnValue` after `state->columns`.


```
GRN_OBJ_FIN(ctx, &(state->columnValue));
GRN_OBJ_FIN(ctx, &(state->columns));
```

Issue #490, opened 2024-06-14
Use Meson as build system


kou:
### What is your problem?

Because PostgreSQL will use Meson as the default build system.

otegami:
I will want to start this issue step by step.
- https://www.youtube.com/watch?v=SAK5RAHwzVQ

otegami:
## TODO

- [ ] handle vendored xxHash
  - https://github.com/pgroonga/pgroonga/pull/750#discussion_r2165171125
- [ ] add test steps
  - https://github.com/pgroonga/pgroonga/pull/750#discussion_r2165188152 
- [ ] add scripts which are used in Make-based build
- [ ] support multi platforms

PR #751, opened 2025-06-25
meson: improve xxHash vendor handling with WrapDB fallback

otegami:
GitHub: GH-490

We replaced xxHash vendor handling with Meson WrapDB support.

I set up subprojects/xxhash.wrap using the following command.

```console
$ meson wrap install xxhash
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -144,6 +144,23 @@ jobs:
         run: |
           cmake --install ..\build
 
+      - name: Install Meson and pkg-config
+        run: |
+          pip install meson ninja
+          choco install pkgconfiglite -y
+      - name: Configure PGroonga by Meson
+        shell: cmd
+        run: |
+          call "%VC_PREFIX%\Auxiliary\Build\vcvars64.bat"
+          set PKG_CONFIG_PATH=..\pgsql\lib\pkgconfig
+          meson setup ..\pgroonga.meson.build ^
+            -Dpg_config="..\pgsql\bin\pg_config.exe"
+      - name: Build PGroonga by Meson
+        shell: cmd
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/windows.yml
```diff
@@ -144,6 +144,23 @@ jobs:
         run: |
           cmake --install ..\build
 
+      - name: Install Meson and pkg-config
+        run: |
+          pip install meson ninja
+          choco install pkgconfiglite -y
```
kou, line 150: We don't need to use `pkg-config` with MSVC.

.github/workflows/windows.yml
```diff
@@ -144,6 +144,23 @@ jobs:
...
+      - name: Configure PGroonga by Meson
+        shell: cmd
+        run: |
+          call "%VC_PREFIX%\Auxiliary\Build\vcvars64.bat"
```
kou, line 154: Do we need this?

Meson can't detect MSVC automatically?

meson.build
```diff
@@ -43,7 +42,12 @@ postgresql = declare_dependency(
   ],
 )
 
-pgrn_dependencies = [postgresql, groonga, libm]
+pgrn_dependencies = [postgresql, groonga]
+
+if host_machine.system() != 'windows'
```
kou, line 47: Should we check MSVC not Windows?

meson.build
```diff
@@ -43,7 +42,12 @@ postgresql = declare_dependency(
...
+  libm = cc.find_library('m', required: false)
```
kou, line 48: Why do we need `required: false`?

otegami:
This change is part of the support for the Windows platform. So I will deal with it at the last step.

otegami:
I will close this PR. I will manage this task in related issue.

PR #758, opened 2025-07-07
Support index_beginscan for PostgreSQL 18 or higher

abetomo:
Arguments have changed since PostgreSQL 18.
https://github.com/postgres/postgres/commit/0fbceae841cb5a31b13d3f284ac8fdd19822eceb

The added an argument is for getting scan statistics and other information.
First, we do not use that information and set the value to NULL, since the behavior will be the same as before.

### DIFF
--- a/src/pgrn-compatible.h
+++ b/src/pgrn-compatible.h
@@ -2,6 +2,8 @@
 
 #include <postgres.h>
 
+#include <access/genam.h>
+
 #ifdef _WIN32
 #	define PRId64 "I64d"
 #	define PRIu64 "I64u"
@@ -73,3 +75,21 @@ typedef Oid PGrnRelFileNumber;
 		((relkind) == RELKIND_RELATION || (relkind) == RELKIND_TOASTVALUE ||   \
 		 (relkind) == RELKIND_MATVIEW)
 #endif
+
+static inline IndexScanDesc
+pgrn_index_beginscan(Relation heapRelation,
+					 Relation indexRelation,
+					 Snapshot snapshot,
+					 int nKeys,
+					 int nOrderBys)
+{
+
+	return index_beginscan(heapRelation,
+						   indexRelation,
+						   snapshot,
+#if PG_VERSION_NUM >= 180000
+						   NULL,
+#endif
+						   nKeys,
+						   nOrderBys);
+}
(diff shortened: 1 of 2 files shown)


abetomo:
Fix the following error:

```
 src/pgrn-query-expand.c:563:73: warning: passing argument 4 of ‘index_beginscan’ makes pointer from integer without a cast [-Wint-conversion]
  563 |                         currentData.table, index, currentData.snapshot, nKeys, nOrderBys);
      |                                                                         ^~~~~
      |                                                                         |
      |                                                                         int
In file included from src/pgrn-query-expand.c:7:
/usr/include/postgresql/18/server/access/genam.h:181:100: note: expected ‘IndexScanInstrumentation *’ but argument is of type ‘int’
  181 |                                                                          IndexScanInstrumentation *instrument,
      |                                                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
src/pgrn-query-expand.c:562:36: error: too few arguments to function ‘index_beginscan’
  562 |                 currentData.scan = index_beginscan(
      |                                    ^~~~~~~~~~~~~~~
/usr/include/postgresql/18/server/access/genam.h:178:22: note: declared here
  178 | extern IndexScanDesc index_beginscan(Relation heapRelation,
      |                      ^~~~~~~~~~~~~~~
```

https://github.com/pgroonga/pgroonga/actions/runs/16106322540/job/45442357009#step:11:60

src/pgrn-compatible.h
```diff
@@ -73,3 +75,21 @@ typedef Oid PGrnRelFileNumber;
...
+					 Snapshot snapshot,
+					 int nKeys,
+					 int nOrderBys)
+{
+
+	return index_beginscan(heapRelation,
+						   indexRelation,
+						   snapshot,
+#if PG_VERSION_NUM >= 180000
+						   NULL,
```
komainu8, line 91: Do you think we should always set NULL?
Do we need to support show index search count?

abetomo: This information is used to display when EXPLAIN ANALYZE.
If we wanted to display additional information there, we could implement it.
First of all, I think the same behavior as before is ok, so I have set NULL.

komainu8: I see.
How about the above comment is written to commit message also.

komainu8 (approved):

PR #762, opened 2025-07-10
Use `pg_noreturn` instead of `pg_attribute_noreturn()`, in PostgreSQL 18

abetomo:
Changed to use `pg_noreturn`.
https://github.com/postgres/postgres/commit/3691edfab97187789b8a1cbb9dce4acf0ecd8f5a

### DIFF
--- a/src/pgroonga-crash-safer.c
+++ b/src/pgroonga-crash-safer.c
@@ -44,6 +44,15 @@ static grn_ctx *ctx = NULL;
 PG_MODULE_MAGIC;
 
 extern PGDLLEXPORT void _PG_init(void);
+
+#if PG_VERSION_NUM >= 180000
+extern PGDLLEXPORT pg_noreturn void
+pgroonga_crash_safer_reset_position_one(Datum datum);
+extern PGDLLEXPORT pg_noreturn void
+pgroonga_crash_safer_reindex_one(Datum datum);
+extern PGDLLEXPORT pg_noreturn void pgroonga_crash_safer_flush_one(Datum datum);
+extern PGDLLEXPORT pg_noreturn void pgroonga_crash_safer_main(Datum datum);
+#else
 extern PGDLLEXPORT void pgroonga_crash_safer_reset_position_one(Datum datum)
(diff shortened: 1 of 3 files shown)


src/pgroonga-crash-safer.c
```diff
@@ -44,6 +44,15 @@ static grn_ctx *ctx = NULL;
 PG_MODULE_MAGIC;
 
 extern PGDLLEXPORT void _PG_init(void);
+
+#if PG_VERSION_NUM >= 180000
```
kou, line 48: We don't want to use these `#if` for this our codes. The approach reduces maintainability. Can we use the following approach?

```diff
diff --git a/src/pgrn-compatible.h b/src/pgrn-compatible.h
index bc3c653d..7dc4e5c7 100644
--- a/src/pgrn-compatible.h
+++ b/src/pgrn-compatible.h
@@ -73,3 +73,9 @@ typedef Oid PGrnRelFileNumber;
 		((relkind) == RELKIND_RELATION || (relkind) == RELKIND_TOASTVALUE ||   \
 		 (relkind) == RELKIND_MATVIEW)
 #endif
+
+#if PG_VERSION_NUM >= 180000
+#	define pg_attribute_noreturn()
+#else
+#	define pg_noreturn
+#endif
diff --git a/src/pgroonga-crash-safer.c b/src/pgroonga-crash-safer.c
index 5350fc14..e659260a 100644
--- a/src/pgroonga-crash-safer.c
+++ b/src/pgroonga-crash-safer.c
@@ -44,7 +44,7 @@ static grn_ctx *ctx = NULL;
 PG_MODULE_MAGIC;
 
 extern PGDLLEXPORT void _PG_init(void);
-extern PGDLLEXPORT void pgroonga_crash_safer_reset_position_one(Datum datum)
+extern PGDLLEXPORT pg_noreturn void pgroonga_crash_safer_reset_position_one(Datum datum)
 	pg_attribute_noreturn();
 extern PGDLLEXPORT void pgroonga_crash_safer_reindex_one(Datum datum)
 	pg_attribute_noreturn();
```

PR #764, opened 2025-07-22
custom-scan: Filter by PGroonga operator

abetomo:

The first goal is to get the results of the following query as expected.

```sql
/* sql/custom-scan/debug.sql */
 SELECT content
   FROM memos
  WHERE content &@~ 'PGroonga OR Groonga';
```

Use PGrnSearchBuildCondition() to filter.

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -158,6 +158,7 @@ PGrnCreateCustomScanState(CustomScan *cscan)
 	state->tableCursor = NULL;
 	GRN_PTR_INIT(&(state->columns), GRN_OBJ_VECTOR, GRN_ID_NIL);
 	GRN_VOID_INIT(&(state->columnValue));
+	memset(&(state->searchData), 0, sizeof(state->searchData));
 	state->searched = NULL;
 
 	return (Node *) &(state->parent);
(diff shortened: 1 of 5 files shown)


otegami (approved):

src/pgrn-custom-scan.c
```diff
@@ -186,31 +189,133 @@ PGrnChooseIndex(Relation table)
...
+	{
+		Expr *expr = (Expr *) lfirst(cell);
+		OpExpr *opexpr;
+		Node *left;
+		Node *right;
+		Var *column;
+		Const *value;
+
+		if (!IsA(expr, OpExpr))
+			continue;
```
kou, line 225: Should we report an error instead of ignore unsupported cases?

src/pgrn-custom-scan.c
```diff
@@ -186,31 +189,133 @@ PGrnChooseIndex(Relation table)
...
 PGrnBeginCustomScan(CustomScanState *customScanState,
 					EState *estate,
 					int eflags)
 {
 	PGrnScanState *state = (PGrnScanState *) customScanState;
-	TupleDesc tupdesc =
-		RelationGetDescr(customScanState->ss.ss_currentRelation);
 	Relation index = PGrnChooseIndex(customScanState->ss.ss_currentRelation);
 	grn_obj *sourcesTable;
+	PGrnSearchData searchData;
```
kou, line 281: We should keep `PGrnSearchData` in `PGrnScanState` to ensure freeing on error.

PR #765, opened 2025-08-04
Optimaize to "match in" from "column &@ keyword1 OR column &@ keyword2 OR ..."

abetomo:
PostgreSQL 18 optimaize to "column &@ ANY(keyword1, keyword2, ...)" from "column &@ keyword1 OR column &@ keyword2 OR ...". 
&@ is match operator. So, we should handle "column &@ keyword1 OR column &@ keyword2 OR ..." as "match in" operator.

### DIFF
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -5391,11 +5391,20 @@ PGrnSearchIsInCondition(ScanKey key)
 			key->sk_strategy == PGrnEqualStrategyNumber);
 }
 
+static bool
+PGrnSearchIsMatchInCondition(ScanKey key)
+{
+	return (key->sk_flags & SK_SEARCHARRAY) &&
+		   ((key->sk_strategy == PGrnMatchStrategyNumber) ||
+			(key->sk_strategy == PGrnMatchStrategyV2Number));
+}
+
 static void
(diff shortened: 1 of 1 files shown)


abetomo:
I've worked on the GH-716 a bit, how about these changes?
Would it be better to add something like `PGrnSearchBuildConditionMatchIn()`?

src/pgroonga.c
```diff
@@ -5391,11 +5391,20 @@ PGrnSearchIsInCondition(ScanKey key)
...
+			(key->sk_strategy == PGrnMatchStrategyV2Number));
+}
+
 static void
 PGrnSearchBuildConditionIn(PGrnSearchData *data,
 						   ScanKey key,
 						   grn_obj *targetColumn,
-						   Form_pg_attribute attribute)
+						   Form_pg_attribute attribute,
+						   bool matchIn)
```
kou, line 5407: If we want to pass something to `PGrnSearchBuildConditionIn()`, we should use `grn_operator op` instead of `bool matchIn`.

If `op` is `GRN_OP_EQUAL`, we can optimize it by `in_values()`.
We should use `column OP value` (such of `GRN_OP_MATCH`) otherwise.

abetomo:
I’ll close this since #716 has started working again.

PR #771, opened 2025-09-11
Add support for amtranslate routines for PostgreSQL 18 or higher

komainu8:
We can implements amtranslatecmptype and amtranslatestrategy in index access method by https://github.com/postgres/postgres/commit/a8025f544854ad8b865c6b4509030ee84aa8f4a0.
If we don't implement them explicitly, PostgreSQL doesn't be regarded as ordered index.

PGroonga can return ordered result. So, we need to implement amtranslatecmptype and amtranslatestrategy.

### DIFF
--- a/src/pgrn-compatible.h
+++ b/src/pgrn-compatible.h
@@ -35,6 +35,10 @@
 #	define PGRN_SUPPORT_PARALLEL_INDEX_BUILD
 #endif
 
+#if PG_VERSION_NUM >= 180000
+#	define PGRN_SUPPORT_AMTRANSLATE_ROUTINE
+#endif
+
 #if PG_VERSION_NUM >= 160000
 #	define PGRN_RELATION_GET_LOCATOR(relation) ((relation)->rd_locator)
 #	define PGRN_RELATION_GET_LOCATOR_NUMBER(relation)                         \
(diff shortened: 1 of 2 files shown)


otegami (approved):

kou:
These failures have gone:

https://github.com/pgroonga/pgroonga/actions/runs/17772908001/job/50513380754#step:11:9

```text
diff -u --color=always /home/runner/work/pgroonga/pgroonga/expected/compare/float4/single/any/equal/empty/indexscan.out /home/runner/work/pgroonga/pgroonga.build/results/compare/float4/single/any/equal/empty/indexscan.out
--- /home/runner/work/pgroonga/pgroonga/expected/compare/float4/single/any/equal/empty/indexscan.out	2025-09-16 16:45:50.216281018 +0000
+++ /home/runner/work/pgroonga/pgroonga.build/results/compare/float4/single/any/equal/empty/indexscan.out	2025-09-16 16:49:16.383021024 +0000
@@ -13,11 +13,13 @@
   FROM ids
  WHERE id = ANY(ARRAY[]::real[])
  ORDER BY id ASC;
-               QUERY PLAN                
------------------------------------------
- Index Scan using pgroonga_index on ids
-   Index Cond: (id = ANY ('{}'::real[]))
-(2 rows)
+                  QUERY PLAN                   
+-----------------------------------------------
+ Sort
+   Sort Key: id
+   ->  Index Scan using pgroonga_index on ids
+         Index Cond: (id = ANY ('{}'::real[]))
+(4 rows)
 
 SELECT id
   FROM ids
```

PR #784, opened 2025-09-22
Fix the FLOAT4OID convert in PGrnConvertToDatum()

abetomo:
The typeID=FLOAT4OID did not convert as expected.
(The value was 0.)
Therefore, we get the value from grn_obj using GRN_FLOAT32_VALUE.

We noticed this while adding a custom scan test, so we'll add that test as well.

### DIFF
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -941,7 +941,13 @@ PGrnConvertToDatum(grn_obj *value, Oid typeID)
 		PG_RETURN_INT64(GRN_INT64_VALUE(value));
 		break;
 	case FLOAT4OID:
-		PG_RETURN_FLOAT4(GRN_FLOAT32_VALUE(value));
+		// For backward compatibility.
+		// `FLOAT4OID` was also `GRN_DB_FLOAT` before.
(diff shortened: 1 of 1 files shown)


kou:
Historically, we use `GRN_DB_FLOAT` for `FLOAT4OID` until 441411db727fa60186e01e3a564eccb6e192bb1a.

Does this change keep backward compatible with indexes that are created before 441411db727fa60186e01e3a564eccb6e192bb1a ?

src/pgroonga.c
```diff
@@ -941,7 +941,7 @@ PGrnConvertToDatum(grn_obj *value, Oid typeID)
 		PG_RETURN_INT64(GRN_INT64_VALUE(value));
 		break;
 	case FLOAT4OID:
-		PG_RETURN_FLOAT4(GRN_FLOAT_VALUE(value));
+		PG_RETURN_FLOAT4(GRN_FLOAT32_VALUE(value));
```
kou, line 944:
```suggestion
		if (value->header.domain == GRN_DB_FLOAT)
			PG_RETURN_FLOAT4(GRN_FLOAT_VALUE(value));
		else
			PG_RETURN_FLOAT4(GRN_FLOAT32_VALUE(value));
```

PR #786, opened 2025-09-24
Add support for `column &= IN ('value1', 'value2', ...)`

komainu8:
PostgreSQL 18 optimizes `column &= 'value1' OR column &= 'value2' OR ...` to `column &= IN ('value1', 'value2', ...)` / `column &= ANY ('{value1,value2,...}'::text[])`.

We need to add support `&=` + `IN`/`ANY` for PostgreSQL 18 support.

Note that `column &= 'value1' OR column &= 'value2' OR ...` with index
scan doesn't use index scan with PostgreSQL 17 or older. (Bitmap scan
can be used.) We don't add support for index scan for `column &=
'value1' OR column &= 'value2' OR ...` with PostgreSQL 17 or older.

We should remove `_{1,2}.out` expected files when we drop support for
PostgreSQL 17 or older. They are only for old PostgreSQL.

### DIFF
--- a/expected/term-search/text/equal/or/indexscan_2.out
+++ b/expected/term-search/text/equal/or/indexscan_2.out
@@ -16,7 +16,7 @@ SELECT name
   FROM tags
  WHERE name = 'groonga' OR name = 'pgroonga'
  ORDER BY id;
-                               QUERY PLAN
+                               QUERY PLAN                                
 -------------------------------------------------------------------------
  Sort
    Sort Key: id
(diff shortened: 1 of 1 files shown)


kou (commented):
Could you move `equal*` in https://github.com/pgroonga/pgroonga/tree/main/sql/term-search/text to `equal*-v2` in a separated PR before this PR?

And then could you use `text/equal{,-v2}/or` not `text/or/equal{,-v2}` in this PR?

PR #789, opened 2025-09-25
Use .editorconfig instead of .dir-locals.el

kou:
GitHub's PR view may use 8 spaces for a tab in `debian/rules` by this.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,8 +1,0 @@
-((c-mode . ((c-file-style . "bsd")
-            (c-basic-offset . 4)
-            (tab-width . 4)
-            (indent-tabs-mode . t)
-            (show-trailing-whitespace . t)))
- (sh-mode . ((indent-tabs-mode . nil)
-             (sh-indentation   . 2)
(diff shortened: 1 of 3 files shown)


kou:
https://github.com/kou/pgroonga/blob/use-editorconfig/packages/postgresql-14-pgdg-pgroonga/debian/rules
This will work.

otegami:
FYI: pre-commit failed on CI.
```
clang-format.............................................................Passed
Lint.....................................................................Passed
meson....................................................................Failed
- hook id: meson-fmt
- exit code: 1

ERROR: Method name must be plain id
[*.{c,h}]
   ^

Error: Process completed with exit code 1.
```

kou:
I've fixed it.

I think that meson-fmt pre-commit configuration should not include `.editorconfig` by default because it's not compatible with Meson format...
https://github.com/trim21/pre-commit-mirror-meson/blob/dd8bbbe2a02c10af0d589dee778d2e584c1911af/.pre-commit-hooks.yaml#L6

otegami (approved):

PR #797, opened 2025-09-26
ci: add PostgreSQL 18

kou:
It's released:
https://www.postgresql.org/about/news/postgresql-18-released-3142/


kou:
The failure is handled as a separated task.

Issue #799, opened 2025-09-26
Use GH_TOKEN in `Helper.detect_latest_groonga_version` if it exists


kou:
### What is your problem?

https://github.com/pgroonga/pgroonga/blob/3650d92406f8532fdf00ff1b5e88f8e884b9df9f/helper.rb#L36-L42

We can avoid the following error by this:

https://github.com/pgroonga/pgroonga/actions/runs/18024009030/job/51288225905?pr=797#step:9:33

```text
OpenURI::HTTPError: 403 rate limit exceeded (OpenURI::HTTPError)
/home/runner/work/pgroonga/pgroonga/helper.rb:38:in 'Helper.detect_latest_groonga_version'
/home/runner/work/pgroonga/pgroonga/packages/pgroonga-package-task.rb:22:in 'GenericPGroongaPackageTask#latest_groonga_version'
/home/runner/work/pgroonga/pgroonga/packages/pgroonga-package-task.rb:77:in 'GenericPGroongaPackageTask#apt_expand_variable'
/home/runner/work/pgroonga/pgroonga/packages/pgroonga-package-task.rb:142:in 'block in VersionedPGroongaPackageTask#apt_prepare_debian_control'
/home/runner/work/pgroonga/pgroonga/apache-arrow/dev/tasks/linux-packages/package-task.rb:106:in 'block in PackageTask#substitute_content'
/home/runner/work/pgroonga/pgroonga/apache-arrow/dev/tasks/linux-packages/package-task.rb:105:in 'String#gsub'
/home/runner/work/pgroonga/pgroonga/apache-arrow/dev/tasks/linux-packages/package-task.rb:105:in 'PackageTask#substitute_content'
/home/runner/work/pgroonga/pgroonga/packages/pgroonga-package-task.rb:141:in 'VersionedPGroongaPackageTask#apt_prepare_debian_control'
/home/runner/work/pgroonga/pgroonga/apache-arrow/dev/tasks/linux-packages/package-task.rb:317:in 'block in PackageTask#apt_prepare_debian_dir'
/home/runner/work/pgroonga/pgroonga/apache-arrow/dev/tasks/linux-packages/package-task.rb:316:in 'IO.open'
/home/runner/work/pgroonga/pgroonga/apache-arrow/dev/tasks/linux-packages/package-task.rb:316:in 'PackageTask#apt_prepare_debian_dir'
[...]

PR #803, opened 2025-09-26
packages: use GH_TOKEN if available

kou:
GitHub: fix GH-799

It's for avoiding rate limit error in CI.


kou:
This broke package CI...

PR #815, opened 2025-09-29
custom-scan: Improve PGrnGetIndexColumnAttributeNumber()

abetomo:
Since IndexInfo has the same value, avoid allocating it multiple times.


abetomo:
Related Comment:
https://github.com/pgroonga/pgroonga/pull/800#discussion_r2383987517

PR #816, opened 2025-09-29
custom-scan: Support multiple indexes

abetomo:

We implemented a simplified logic to choose among multiple indexes.

PGrnChooseIndex() function was executed twice during the creation of the plan path and the preparation for the scan.
We change it to execute only once.

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -26,7 +26,7 @@ typedef struct PGrnScanIndexData
 {
 	Oid indexOID;
 	ScanKeyData *scanKeys;
-	int nScankeys;
+	int nScanKeys;
 } PGrnScanIndexData;
 
 typedef struct PGrnScanState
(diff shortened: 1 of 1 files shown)


src/pgrn-custom-scan.c
```diff
@@ -22,9 +22,17 @@
 #include "pgrn-groonga.h"
 #include "pgrn-search.h"
 
+typedef struct PGrnScanIndexData
+{
+	Oid indexOid;
```
kou, line 27:
```suggestion
	Oid indexOID;
```

src/pgrn-custom-scan.c
```diff
@@ -22,9 +22,17 @@
...
+	ScanKeyData *scankeys;
+	int scankeySize;
```
kou, line 29:
```suggestion
	int nScanKeys;
```

src/pgrn-custom-scan.c
```diff
@@ -22,9 +22,17 @@
 #include "pgrn-groonga.h"
 #include "pgrn-search.h"
 
+typedef struct PGrnScanIndexData
+{
+	Oid indexOID;
+	ScanKeyData *scanKeys;
+	int nScankeys;
```
kou, line 29:
```suggestion
	int nScanKeys;
```

PR #819, opened 2025-10-01
custom-scan: Fix getting PGrnSetTargetColumns's column name

abetomo:
Fix cases where entry->resname was not set.

An issue found while verifying whether `sql/full-text-search/text/single/declarative-partitioning.sql` query works with custom scan. 
The test for that was also added.


sql/full-text-search/text/single/partitioning/declarative-partitioning.sql
abetomo, line 1: Moved it when adding the test for custom scan.

kou: Ah, we created `partitioning/` directory and this was moved to it?

abetomo: I didn't explain it well enough.
Yes, that's right.

abetomo: That's sounds good.

PR #823, opened 2025-10-01
Revert "varchar/query-v2: Fix many-partition test"

abetomo:
Reverts pgroonga/pgroonga#820

The test failed on Windows.


abetomo:
Thanks for the fix.

PR #827, opened 2025-10-03
custom-scan: No custom scan for unsupported operators

abetomo:

For unsupported operators, a standard scan will be executed.


kou:
This is OK for now. But we should evaluate the sequential scan in Groonga for faster evaluation than PosgreSQL level sequential scan in the future. (It's low priority.)

PR #829, opened 2025-10-03
custom-scan: Prepare to support parallel scan

abetomo:
Added the necessary functions. First, add them as empty:
* PGrnEstimateDSMCustomScan
* PGrnInitializeDSMCustomScan
* PGrnReInitializeDSMCustomScan
* PGrnInitializeWorkerCustomScan

We also added a `NodeTag` at the head of `PGrnScanIndexData`.
`PGrnScanIndexData` will be added to `custom_private`, but the elements added to `custom_private` must be `Node`.
Without a `NodeTag`, the following error occurs when running in parallel.

```
ERROR:  unrecognized node type: XXX
```

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -24,6 +24,7 @@
 
 typedef struct PGrnScanIndexData
 {
+	NodeTag type;
 	Oid indexOID;
 	ScanKeyData *scanKeys;
 	int nScanKeys;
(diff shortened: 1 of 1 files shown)


src/pgrn-custom-scan.c
```diff
@@ -292,6 +310,7 @@ PGrnSetRelPathlistHook(PlannerInfo *root,
 			scanData = palloc(sizeof(PGrnScanIndexData));
 			scanData->scanKeys =
 				palloc(sizeof(ScanKeyData) * list_length(quals));
+			scanData->type = T_CustomScan;
```
abetomo, line 313: I arbitrarily set `T_CustomScan` as the `NodeTag`, but since this is not appropriate, I will reconsider `PGrnScanIndexData` itself.

PR #830, opened 2025-10-03
custom-scan: Set the value of `custom_private` to `Node` only

abetomo:
We need to set only `Node` to `custom_private`, especially when using parallel scan.
In the code before the modification, `PGrnScanIndexData` and `ScanKeyData` are not `Node`, so an error occurs during parallel scan.

Terefore, we replaced them with `Node` type provided by PostgreSQL, such as `List`.

Although it is possible to make `PGrnScanIndexData` `Node`, it is simpler to use the types provided by PostgreSQL.

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -147,6 +147,11 @@ PGrnScanKeySourceMake(int indexAttributeNumber,
 					  Oid opFuncID,
 					  Const *value)
 {
+	/*
+	 * ScanKeySource will be set to `custom_private`.
+	 * Since only a `Node` can be set to `custom_private`, a `List`, which is a
+	 * `Node`, will be used to create the value.
+	 */
 	return list_make3(list_make2_int(indexAttributeNumber, indexStrategy),
 					  list_make1_oid(opFuncID),
 					  list_make1(value));
@@ -179,6 +184,8 @@ PGrnScanKeySourceGetValue(List *source)
 static List *
 PGrnCustomPrivateMake(Oid indexOID, List *scanKeySources)
 {
+	// Only a `Node` can be set to `custom_private`.
+	// See also the comments in PGrnScanKeySourceMake().
 	return list_make2(list_make1_oid(indexOID), scanKeySources);
 }
 


src/pgrn-custom-scan.c
```diff
@@ -377,7 +369,8 @@ PGrnCreateCustomScanState(CustomScan *cscan)
 	memset(&(state->searchData), 0, sizeof(state->searchData));
 	state->searched = NULL;
 	state->scoreAccessor = NULL;
-	state->scanData = linitial(cscan->custom_private);
+	state->indexOID = linitial_oid(linitial(cscan->custom_private));
+	state->scanKeySources = lsecond(cscan->custom_private);
```
kou, line 373: ditto.

src/pgrn-custom-scan.c
```diff
@@ -148,14 +142,12 @@ PGrnOpInOpfamily(Relation index, OpExpr *opexpr)
 }
 
 static void
-PGrnScanIndexDataInit(Relation index, List *quals, PGrnScanIndexData *data)
+PGrnTraverseQuals(Relation index, List *quals, List **scanKeySources)
```
kou, line 145: Can we use better name?
This doesn't just traverse quals, right?
This collects scan key sources from quals, right?

src/pgrn-custom-scan.c
```diff
@@ -148,14 +142,12 @@ PGrnOpInOpfamily(Relation index, OpExpr *opexpr)
...
 {
 	const char *tag = "pgroonga: [custom-scan][index-data][init]";
```
kou, line 147: Could you update this too?

src/pgrn-custom-scan.c
```diff
@@ -265,8 +257,8 @@ PGrnChooseIndex(Relation table, List *quals, PGrnScanIndexData *data)
 			RelationClose(index);
 			continue;
 		}
-		PGrnScanIndexDataInit(index, quals, data);
-		if (data->nScanKeys == 0)
+		PGrnTraverseQuals(index, quals, scanKeySources);
+		if (*scanKeySources == NIL)
```
kou, line 261:
```suggestion
		if (!*scanKeySources)
```

src/pgrn-custom-scan.c
```diff
@@ -309,15 +302,13 @@ PGrnSetRelPathlistHook(PlannerInfo *root,
 		{
 			Relation index;
 			List *quals = PGrnConvertExprList(rel->baserestrictinfo);
-			scanData = palloc(sizeof(PGrnScanIndexData));
-			scanData->scanKeys =
-				palloc(sizeof(ScanKeyData) * list_length(quals));
-			index = PGrnChooseIndex(table, quals, scanData);
+			index = PGrnChooseIndex(table, quals, &scanKeySources);
```
kou, line 305: Can we use `custom_private = PGrnChooseIndex(table, quals)`?

src/pgrn-custom-scan.c
```diff
@@ -233,48 +279,48 @@ PGrnScanIndexDataInit(Relation index, List *quals, PGrnScanIndexData *data)
...
+		scanKeySources = PGrnCollectScanKeySources(index, quals);
+		if (!scanKeySources)
 		{
 			RelationClose(index);
 			continue;
 		}
-
-		return index;
+		RelationClose(index);
+		return PGrnCustomPrivateMake(indexOID, scanKeySources);
```
kou, line 321:
```c
RelationClose(index);
if (!scanKeySources)
    continue;
return PGrnCustomPrivateMake(indexOID, scanKeySources);
```

kou (approved):
Could you add a comment why we need to use `List` as a comment before you merge this?

PR #832, opened 2025-10-05
custom-scan: Support sorting by index

abetomo:
By setting `cpath->path.pathkeys`, we inform the planner that the custom scan results are sorted.

`pathkeys` is not included in `CustomScanState`, so we use `custom_private` to pass them.

### DIFF
--- a/src/pgrn-compatible.h
+++ b/src/pgrn-compatible.h
@@ -95,6 +95,12 @@ typedef Oid PGrnRelFileNumber;
 #	define pg_noreturn
 #endif
 
+#if PG_VERSION_NUM < 180000
+#	define COMPARE_LT BTLessStrategyNumber
+#	define COMPARE_GT BTGreaterStrategyNumber
+#	define pk_cmptype pk_strategy
+#endif
+
 static inline IndexScanDesc
 pgrn_index_beginscan(Relation heapRelation,
 					 Relation indexRelation,
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -27,11 +27,13 @@ typedef struct PGrnScanState
 	CustomScanState parent; /* must be first field */
 	Oid indexOID;
 	List *scanKeySources;
+	List *pathKeys;
 	grn_table_cursor *tableCursor;
 	grn_obj columns;
 	grn_obj columnValue;
 	PGrnSearchData searchData;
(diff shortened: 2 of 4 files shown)


src/pgrn-compatible.h
```diff
@@ -95,6 +95,12 @@ typedef Oid PGrnRelFileNumber;
 #	define pg_noreturn
 #endif
 
+#if PG_VERSION_NUM < 180000
+#	define COMPARE_LT BTLessStrategyNumber
+#	define COMPARE_GT BTGreaterStrategyNumber
+#	define pk_cmptype pk_strategy
```
abetomo, line 101: The name has changed from PostgreSQL18.
https://github.com/postgres/postgres/commit/8123e91f

src/pgrn-custom-scan.c
```diff
@@ -352,14 +360,45 @@ PGrnCollectScanKeySources(Relation index, List *quals)
...
+{
+	TupleDesc tupdesc = RelationGetDescr(index);
+	for (AttrNumber i = 0; i < tupdesc->natts; i++)
+	{
+		Form_pg_attribute attr = TupleDescAttr(tupdesc, i);
+		if (strcmp(NameStr(attr->attname), name) == 0)
+			return true;
+	}
+	return false;
+}
```
abetomo, line 374: PGrnIndexContainColumn:
I just changed the location of the existing function definition.

src/pgrn-custom-scan.c
```diff
@@ -570,6 +610,48 @@ PGrnSearchBuildCustomScanConditions(CustomScanState *customScanState,
...
+		}
+	}
+
+	state->sorted = grn_table_create(
+		ctx, NULL, 0, NULL, GRN_OBJ_TABLE_NO_KEY, NULL, state->searched);
+	grn_table_sort(
+		ctx, state->searched, 0, -1, state->sorted, sortKeys, nSortKeys);
+
+	for (i = 0; i < nSortKeys; i++)
+		grn_obj_close(ctx, sortKeys[i].key);
```
abetomo, line 651: Since it was allocated in `(grn_table_sort_key *) palloc(sizeof(grn_table_sort_key) * nSortKeys)`, `grn_table_sort_key_close` could not be used.
Is there a better way?

abetomo:
It will be closed due to an implementation error.

PR #834, opened 2025-10-05
custom-scan: Disable registered functions

abetomo:
Since there is no UnregisterCustomScanMethods() that is paired with RegisterCustomScanMethods(), we added a `PGrnCustomScanInitialized` flag, and when the flag is disabled, each function does nothing.

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -337,9 +337,6 @@ PGrnSetRelPathlistHook(PlannerInfo *root,
 	CustomPath *cpath;
 	List *privateData = NIL;
 
-	if (!PGrnCustomScanInitialized)
-		return;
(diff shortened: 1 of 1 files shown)


src/pgrn-custom-scan.c
```diff
@@ -392,6 +337,9 @@ PGrnSetRelPathlistHook(PlannerInfo *root,
 	CustomPath *cpath;
 	List *privateData = NIL;
 
+	if (!PGrnCustomScanInitialized)
+		return;
+
```
kou, line 342: Do we need this?
It seems that this isn't called after `PGrnFinalizeCustomScan()`.

abetomo: That’s true.

PR #835, opened 2025-10-06
custom-scan: Support sorting by index

abetomo:
By setting `cpath->path.pathkeys`, we inform the planner that the custom scan results are sorted.

`pathkeys` is not included in `CustomScanState`, so we use `custom_private` to pass them.

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -8,6 +8,7 @@
 #include <executor/executor.h>
 #include <nodes/extensible.h>
 #include <nodes/nodeFuncs.h>
+#include <optimizer/optimizer.h>
 #include <optimizer/pathnode.h>
 #include <optimizer/paths.h>
 #include <optimizer/restrictinfo.h>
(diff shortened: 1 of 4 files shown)


src/pgrn-custom-scan.c
```diff
@@ -593,6 +643,54 @@ PGrnSearchBuildCustomScanConditions(CustomScanState *customScanState,
...
+		}
+	}
+
+	state->sorted = grn_table_create(
+		ctx, NULL, 0, NULL, GRN_OBJ_TABLE_NO_KEY, NULL, state->searched);
+	grn_table_sort(
+		ctx, state->searched, 0, -1, state->sorted, sortKeys, nSortKeys);
+
+	for (unsigned int i = 0; i < nSortKeys; i++)
+		grn_obj_close(ctx, sortKeys[i].key);
```
kou, line 690: We can use `grn_obj_close()` here because all `.key`s are accessor (`state->searched` doesn't have any column) but, in general, we should use `grn_obj_unlink()` for `grn_obj_column()`. `grn_obj_unlink()` closes an accessor but doesn't close a column. 

```suggestion
		grn_obj_unlink(ctx, sortKeys[i].key);
```

src/pgrn-custom-scan.c
```diff
@@ -827,6 +924,11 @@ PGrnEndCustomScan(CustomScanState *customScanState)
 		grn_obj_close(ctx, state->searched);
 		state->searched = NULL;
 	}
+	if (state->sorted)
+	{
+		grn_obj_close(ctx, state->sorted);
+		state->sorted = NULL;
+	}
```
kou, line 931: We should close `sorted` AFTER `searched` because `searched` is referred from `sorted`.

PR #838, opened 2025-10-06
custom-scan: Refactoring

abetomo:
* Unify the argument name of CustomScanState to customScanState.
* Create a function that sets ResultTupleSlot in ExecCustomScan.
  * There are no changes to the logic.
  * This is preparation for changing ExecCustomScan later.
* Add initialization of state->ctidAccessor.
  * This is more of an addition to fix an omission rather than refactoring.

### DIFF
--- a/src/pgrn-custom-scan.c
+++ b/src/pgrn-custom-scan.c
@@ -768,9 +768,9 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
 }
 
 static TupleTableSlot *
-PGrnResultTupleSlot(CustomScanState *customScanState,
-					grn_id id,
-					ItemPointerData ctid)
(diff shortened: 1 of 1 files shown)


src/pgrn-custom-scan.c
```diff
@@ -764,6 +767,97 @@ PGrnBeginCustomScan(CustomScanState *customScanState,
 	memset(&(state->searchData), 0, sizeof(state->searchData));
 }
 
+static TupleTableSlot *
+PGrnResultTupleSlot(CustomScanState *customScanState,
```
kou, line 771: Could you use verb for function name?
`PGrnMakeTupleTableSlot()`? `PGrnBuildTupleTableSlot()`?

PR #876, opened 2025-10-24
test: revert GRN_LANGUAGE_MODEL_DOWNLOAD_CACHE_DIR in test

kou:
`\getenv` gets a client side environment variable not a server side's one. So we don't need to use the same variable name and same variable value.


abetomo:
To run tests in CI, we need to set PGRN_LANGUAGE_MODEL_TEST in CI.

Issue #881, opened 2025-10-26
ci: PGDG + AlmaLinux 10 jobs are failing


kou:
### What happened?

https://github.com/pgroonga/pgroonga/actions/runs/18814420816/job/53680593884?pr=880#step:14:349

```text
    Running scriptlet: pgbackrest-2.57.0-1PGDG.rhel10.x86_64                 5/68 
  Creating group 'postgres' with GID 26.
  Creating user 'postgres' (PostgreSQL Server) with UID 26 and GID 26.
  /usr/bin/chown: cannot access '/var/lib/pgsql': No such file or directory
  error: %prein(pgbackrest-2.57.0-1PGDG.rhel10.x86_64) scriptlet failed, exit status 1
  
  Error in PREIN scriptlet in rpm package pgbackrest
    Running scriptlet: postgresql15-server-15.14-1PGDG.rhel10.x86_64         6/68 
  error: pgbackrest-2.57.0-1PGDG.rhel10.x86_64: install failed
```

It seems that this is a pgbackrest package problem. It requires `/var/lib/pgsql` but it doesn't exist yet.

### How to reproduce it

Run CI.

### Expected behavior

No error.

### OS and PostgreSQL version

- OS: AlmaLinux
- OS Version: 10
- PostgreSQL Version: Not related

### PGroonga version

Not related

### Groonga version

Not related

komainu8:
I reported upstream at https://github.com/pgdg-packaging/pgdg-rpms/issues/105 and I implement work around of this issues at https://github.com/pgroonga/pgroonga/pull/883.

PR #883, opened 2025-10-27
ci: workaround for dependency package install failure

komainu8:
Fix: GH-881

This is workaround.

Currently, installation of pgbackrest package is failure cause of https://github.com/pgdg-packaging/pgdg-rpms/issues/105.
postgresql*-contrib weak dependency pgbackrest.
But even if pgbackrest does not install, have no impact on executing PGroonga's tests.

Therefore, we disable EPEL repository when install postgresql*-contrib.
The pgbackrest package depends on libssh2, which is only available from the EPEL repository.
As a result, pgbackrest is installed only when EPEL is enabled.

If we disable EPEL repository, have no impact on installing postgresql*-contrib.
Because The installation of postgresql*-contrib use EPEL repository only when pgbackrest installed. 
Also, postgresql*-contrib works correctly even if pgbackrest does not installed.
because pgbackrest is weak dependency.

Tests are all passed by this work around as below.
https://github.com/komainu8/pgroonga/actions/runs/18830781666/job/53726404046

### DIFF
--- a/packages/yum/test.sh
+++ b/packages/yum/test.sh
@@ -91,7 +91,7 @@ pgroonga_package=$(basename $(ls ${packages_dir}/*-pgroonga-*.rpm | head -n1) | 
                      sed -e 's/-pgroonga-.*$/-pgroonga/g')
 postgresql_version=$(echo ${pgroonga_package} | grep -E -o '[0-9.]+')
 
-# This condition is workaround.
+# TODO: This condition is workaround.
 # We can remove when the following issue resolved.
 # https://github.com/pgdg-packaging/pgdg-rpms/issues/105
 if [ "${os}" = "almalinux" ] && [ ${major_version} = "10" ]; then


packages/yum/test.sh
```diff
@@ -91,7 +91,12 @@ pgroonga_package=$(basename $(ls ${packages_dir}/*-pgroonga-*.rpm | head -n1) |
                      sed -e 's/-pgroonga-.*$/-pgroonga/g')
 postgresql_version=$(echo ${pgroonga_package} | grep -E -o '[0-9.]+')
 
-${DNF} install -y postgresql${postgresql_version}-contrib
+if [ "${os}" = "almalinux" ] || [ ${major_version} = "10" ]; then
```
kou: Could you add a comment when we can remove this workaround?

komainu8: I added a comment.

packages/yum/test.sh
```diff
@@ -91,7 +91,15 @@ pgroonga_package=$(basename $(ls ${packages_dir}/*-pgroonga-*.rpm | head -n1) |
                      sed -e 's/-pgroonga-.*$/-pgroonga/g')
 postgresql_version=$(echo ${pgroonga_package} | grep -E -o '[0-9.]+')
 
-${DNF} install -y postgresql${postgresql_version}-contrib
+# This condition is workaround.
```
otegami: How about adding a TODO prefix so that we can easily grep this line?
```suggestion
# TODO: This condition is workaround.
```

komainu8: I added it.

PR #900, opened 2025-11-19
language-model: add support for prefix

komainu8:
Support the `passage_prefix` and `query_prefix` options supported by Groonga.
https://github.com/groonga/groonga/pull/2628

### DIFF
--- a/expected/semantic-search/text/similar-distance-condition-v2/with-condition.out
+++ b/expected/semantic-search/text/similar-distance-condition-v2/with-condition.out
@@ -14,7 +14,9 @@ INSERT INTO memos VALUES (3, 'Ruby is a object oriented script language.');
 CREATE INDEX pgrn_index ON memos
  USING pgroonga (id, content pgroonga_text_semantic_search_ops_v2)
  WITH (plugins = 'language_model/knn',
-       model = 'hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF');
+       model = 'hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF',
+       passage_prefix = 'passage: ',
+       query_prefix = 'query: ');
 SET enable_seqscan = off;
 SET enable_indexscan = on;
 SET enable_bitmapscan = off;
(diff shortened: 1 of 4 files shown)


src/pgrn-options.c
```diff
@@ -560,6 +562,18 @@ PGrnInitializeOptions(void)
...
 					  lock_mode);
+	add_string_reloption(PGrnReloptionKind,
+						 "passage_prefix",
+						 "Specify prefix for search target text",
+						 NULL,
+						 false,
+						 lock_mode);
+	add_string_reloption(PGrnReloptionKind,
+						 "query_prefix",
+						 "Specify  prefix for query text",
```
kou:
```suggestion
						 "Specify prefix for query text",
```

expected/semantic-search/text/similar-condition-v2/options/prefix.out
```diff
@@ -0,0 +1,47 @@
...
+);
+INSERT INTO memos VALUES (1, 'PostgreSQL is a RDBMS.');
+INSERT INTO memos VALUES (2, 'Groonga is fast full text search engine.');
+INSERT INTO memos VALUES (3, 'PGroonga is a PostgreSQL extension that uses Groonga.');
+CREATE INDEX pgrn_index ON memos
+ USING pgroonga (content pgroonga_text_semantic_search_ops_v2)
+ WITH (plugins = 'language_model/knn',
+       model = 'hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF',
+       passage_prefix = 'passage: ',
+       query_prefix = 'query: ');
```
kou: Could you also use them in https://github.com/pgroonga/pgroonga/blob/main/sql/semantic-search/text/similar-distance-condition-v2/with-condition.sql and https://github.com/pgroonga/pgroonga/blob/main/sql/semantic-search/text/similar-distance-condition-v2/without-condition.sql ?

We should use them with `groonga/multilingual-e5-base-Q4_K_M-GGUF`.

komainu8:
+1

Issue #906, opened 2025-11-27
`pgroonga_language_model_vectorize()`: mark it as `STABLE` or `IMMUTABLE`


kou:
### What is your problem?

If we can mark `pgroonga_language_model_vectorize()` as `STABLE` or `IMMUTABLE`, PostgreSQL can cache the result and avoid needless function calls. `pgroonga_language_model_vectorize()` is a heavy process. So it will improve performance.

See also: https://www.postgresql.org/docs/current/xfunc-volatility.html

PR #909, opened 2025-12-01
language_model_vectorize: Add `IMMUTABLE` to `CREATE FUNCTION`

abetomo:
GitHub fixes GH-906

In reality, pgroonga_language_model_vectorize() is not `IMMUTABLE`.
But, due to the nature of the function, some minor differences are acceptable, and for performance reasons, it is marked as `IMMUTABLE`.


kou:
Could you try with `FROM normal_table`?

kou:
Hmm... Let's use `IMMUTABLE`...

PR #911, opened 2025-12-02
language_model_vectorize: fix a bug that causes currentModelName to be invalid

abetomo:
The process may crash when executing the following query multiple times with pgvector.

```
SELECT pgroonga_language_model_vectorize(
  'hf:///groonga/all-MiniLM-L6-v2-Q4_K_M-GGUF',
  column)::vector;
```

This is caused by `currentModelName` space being reused during the process, so we store it in `grn_obj` to persist it.

### DIFF
--- a/src/pgrn-language-model-vectorize.c
+++ b/src/pgrn-language-model-vectorize.c
@@ -23,7 +23,7 @@ PGrnInitializeLanguageModelVectorize(void)
 {
 	loader = grn_language_model_loader_open(ctx);
 	GRN_FLOAT32_INIT(&vector, GRN_OBJ_VECTOR);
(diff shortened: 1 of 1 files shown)


src/pgrn-language-model-vectorize.c
```diff
@@ -23,6 +23,7 @@ PGrnInitializeLanguageModelVectorize(void)
 {
 	loader = grn_language_model_loader_open(ctx);
 	GRN_FLOAT32_INIT(&vector, GRN_OBJ_VECTOR);
+	GRN_TEXT_INIT(&currentModelName, GRN_OBJ_DO_SHALLOW_COPY);
```
kou: We can't use `GRN_OBJ_DO_SHALLOW_COPY`. If we use it, `modelName` isn't copied.

PR #917, opened 2025-12-10
Add support for "column LIKE/ILIKE ANY ('{keyword1,keyword2,...}'::text[])"

komainu8:
Fix: GH-916

This is the same issue as GH-716.

PostgreSQL 18 optimizes `column LIKE/ILIKE keyword1 OR column LIKE/ILIKE keyword2 OR ...` to `column LIKE/ILIKE ANY ('{keyword1,keyword2,...}'::text[])`.

We need to add support for `LIKE/ILIKE` + `ANY` for PostgreSQL 18 support.

Note that `column LIKE/ILIKE keyword1 OR column LIKE/ILIKE keyword2 OR ...` with index scan doesn't use index scan with PostgreSQL 17 or older. 
(Bitmap scan can be used.) We don't add support for index scan for `column LIKE/ILIKE keyword1 OR column LIKE/ILIKE keyword2 OR ...` with PostgreSQL 17 or older.

Reported by kurita0. Thanks!!!

### DIFF
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -5489,9 +5489,21 @@ PGrnSearchIsMatchInCondition(ScanKey key)
 {
 	return (key->sk_flags & SK_SEARCHARRAY) &&
 		   ((key->sk_strategy == PGrnMatchStrategyNumber) ||
-			(key->sk_strategy == PGrnMatchStrategyV2Number) ||
-			(key->sk_strategy == PGrnLikeStrategyNumber) ||
-			(key->sk_strategy == PGrnILikeStrategyNumber));
+			(key->sk_strategy == PGrnMatchStrategyV2Number));
+}
+
+static bool
+PGrnSearchIsLikeAnyCondition(ScanKey key)
+{
+	return (key->sk_flags & SK_SEARCHARRAY) &&
+		   (key->sk_strategy == PGrnLikeStrategyNumber);
+}
+
+static bool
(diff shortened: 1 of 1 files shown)


src/pgroonga.c
```diff
@@ -5489,9 +5489,15 @@ PGrnSearchIsMatchInCondition(ScanKey key)
 {
 	return (key->sk_flags & SK_SEARCHARRAY) &&
 		   ((key->sk_strategy == PGrnMatchStrategyNumber) ||
-			(key->sk_strategy == PGrnMatchStrategyV2Number));
+			(key->sk_strategy == PGrnMatchStrategyV2Number) ||
+			(key->sk_strategy == PGrnLikeStrategyNumber) ||
+			(key->sk_strategy == PGrnILikeStrategyNumber));
 }
```
kou: Please don't reuse this function. This is for match condition (`&@  (...)`) not `LIKE`/`ILIKE`.

BTW, we should use `Any` not `In` for this function name... Because `COLUMN OPERATOR IN (...)` is invalid syntax. `COLUMN OPERATOR ANY (...)` is valid.

komainu8: I don't reuse this function. I created functions for `LIKE` and `ILIKE` at https://github.com/pgroonga/pgroonga/pull/917/commits/afe2f5d98d3138c3d50c7586e1dda558d1715be8.

Modify of this function name  will be fixed in a separate task.

PR #918, opened 2025-12-11
Fix build error on PostgreSQL19

komainu8:
We will follow the next changes in PostgreSQL:
https://github.com/postgres/postgres/commit/0986e95161cec929d8f39c01e9848f34526be421

In PostgreSQL 19 and later, use `JsonbInState` instead of `JsonbParseState`.

### DIFF
--- a/src/pgrn-result-converter.c
+++ b/src/pgrn-result-converter.c
@@ -659,7 +659,5 @@ PGrnResultConverterBuildJSONBObjects(PGrnResultConverter *converter)
 		pushJsonbValue(&state, WJB_END_OBJECT, NULL);
 	}
 
-	{
-		PGrnJsonbValueToJsonb(state);
-	}
+	return PGrnJsonbValueToJsonb(state);
 }


komainu8 (commented):
+1

src/pgrn-compatible.h
```diff
@@ -101,6 +101,13 @@ typedef Oid PGrnRelFileNumber;
 #	define pk_cmptype pk_strategy
 #endif
 
+#include <utils/jsonb.h>
+#if PG_VERSION_NUM >= 190000
+typedef JsonbInState PGrnJsonState;
+#else
+typedef JsonbParseState *PGrnJsonState;
+#endif
```
otegami:
```suggestion
#include <utils/jsonb.h>
#if PG_VERSION_NUM >= 190000
	typedef JsonbInState PGrnJsonState;
#else
	typedef JsonbParseState *PGrnJsonState;
#endif
```

PR #953, opened 2026-05-12
packages ubuntu: add support for Ubuntu 26.04 with PostgreSQL for Ubuntu official package

komainu8:
This modification only support Ubuntu 26.04 LTS with PostgreSQL for Ubuntu official package( https://packages.ubuntu.com/resolute/postgresql-18 ).


packages/apt/test.sh
```diff
@@ -80,12 +80,6 @@ wget \
   https://packages.groonga.org/${os}/groonga-apt-source-latest-${code_name}.deb
 apt install -V -y ./groonga-apt-source-latest-${code_name}.deb
 
-if [ "${os}" = "ubuntu" ]; then
-  apt install -y software-properties-common
-  add-apt-repository -y universe
-  add-apt-repository -y ppa:groonga/ppa
-fi
-
```
komainu8, line 88: We should use Groonga APT repository ([packages.groonga.org](http://packages.groonga.org/)) instead of  
The Grppmga PPA (Personal Package Archive, ppa:groonga/ppa) in test.

Because packages from that Groonga APT repository are built with Apache Arrow enabled. 

So, I removed these codes.

PR #974, opened 2026-06-23
Fix a crash when an UPDATE statement with a WHERE clause and VACUUM are executed concurrently

komainu8:
We should not call EnsureLatestDB() between pgroonga_beginscan() and pgroonga_endscan() according to the comment for PGrnEnsureLatestDB() in https://github.com/pgroonga/pgroonga/blob/4.0.6/src/pgroonga.c#L792-L814.
In fact, a PGroonga process does not detect a VACUUM executed by another process and does not call EnsureLatestDB() in this scope.

However, an UPDATE statement with a WHERE clause may call pgroonga_insert() in this scope as below:

```
pgroonga: [trace][pgroonga_beginscan][enter]
pgroonga: [trace][pgroonga_beginscan][exit]
pgroonga: [trace][pgroonga_rescan][enter]
pgroonga: [trace][pgroonga_rescan][exit]
pgroonga: [trace][pgroonga_insert][enter]
pgroonga: [trace][PGrnEnsureLatestDB][enter]
pgroonga: [trace][PGrnEnsureLatestDB][exit]
pgroonga: [trace][pgroonga_insert][exit]
pgroonga: [trace][pgroonga_endscan][enter]
pgroonga: [trace][pgroonga_endscan][exit]
```

pgroonga_insert() detects execution of VACUUM and close the opened grn_obj by calling PGrnUnmapDB().
Therefore, if an UPDATE statement with a WHERE clause and VACUUM are executed concurrently, PGronga close the opened grn_obj between pgroonga_beginscan() and pgroonga_endscan() as below:

```
pgroonga: [trace][pgroonga_beginscan][enter]
pgroonga: [trace][pgroonga_beginscan][exit]
pgroonga: [trace][pgroonga_rescan][enter]
pgroonga: [trace][pgroonga_rescan][exit]
pgroonga: [trace][pgroonga_insert][enter]
pgroonga: [trace][PGrnEnsureLatestDB][enter]
pgroonga: [trace][PGrnUnmapDB][enter]
pgroonga: [trace][PGrnUnmapDB][exit]
pgroonga: [trace][PGrnEnsureLatestDB][exit]
pgroonga: [trace][pgroonga_insert][exit]
pgroonga: [trace][pgroonga_endscan][enter]
pgroonga: [trace][pgroonga_endscan][exit]
```

We add a condition to prevent PGrnEnsureLatestDB() from being executed while between pgroonga_beginscan() and pgroonga_endscan in this modification.
With the change, PGroonga will no longer crash even in the scenario described above.

### DIFF
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -830,12 +830,6 @@ PGrnEnsureLatestDB(void)
 		return false;
 	}
 
-	if (PGrnNScanOpaques != 0)
-	{
-		PGRN_TRACE_LOG_EXIT();
-		return false;
-	}
-
 	GRN_LOG(
 		ctx, GRN_LOG_DEBUG, "pgroonga: unmap DB because VACUUM was executed");
 	PGrnUnmapDB();
@@ -5190,7 +5184,10 @@ pgroonga_insert(Relation index,
 						tag)));
 	}
 
-	PGrnEnsureLatestDB();
+	if (PGrnNScanOpaques == 0)
+	{
+		PGrnEnsureLatestDB();
+	}
 
 	PGrnWALApply(index);
 
--- a/src/pgroonga.c
+++ b/src/pgroonga.c
@@ -100,11 +100,11 @@ static bool PGrnBaseInitialized = false;
 bool PGrnGroongaInitialized = false;
 static bool PGrnCrashSaferInitialized = false;
 bool PGrnEnableParallelBuildCopy = false;
+static bool PGrnIsScanning = false;
 
 typedef struct PGrnProcessSharedData
 {
 	TimestampTz lastVacuumTimestamp;
-	pg_atomic_uint32 nScanningProcesses;
 } PGrnProcessSharedData;
 
 typedef struct PGrnProcessLocalData
@@ -729,8 +729,8 @@ _PG_init(void)
 			if (!found)
 			{
 				processSharedData->lastVacuumTimestamp = GetCurrentTimestamp();
-				pg_atomic_init_u32(&(processSharedData->nScanningProcesses), 0);
 			}
+			PGrnIsScanning = false;
 			LWLockRelease(AddinShmemInitLock);
 		}
 		processLocalData.lastDBUnmapTimestamp = GetCurrentTimestamp();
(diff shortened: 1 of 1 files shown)


src/pgroonga.c
```diff
@@ -103,6 +104,7 @@ bool PGrnEnableParallelBuildCopy = false;
 typedef struct PGrnProcessSharedData
 {
 	TimestampTz lastVacuumTimestamp;
+	pg_atomic_uint32 nScanningProcesses;
```
kou: Why do we need to make this process shared?

src/pgroonga.c
```diff
@@ -99,6 +100,7 @@ static bool PGrnBaseInitialized = false;
 bool PGrnGroongaInitialized = false;
 static bool PGrnCrashSaferInitialized = false;
 bool PGrnEnableParallelBuildCopy = false;
+static bool PGrnIsScanning = false;
```
kou: Do we really need to introduce a new variable?

Can we reuse `PGrnNScanOpaques`?

komainu8: Ah. I see. I had not noticed that. I'll check it.

src/pgroonga.c
```diff
@@ -79,6 +79,7 @@
 #include <groonga.h>
 
 #include <math.h>
+#include <port/atomics.h>
```
kou: Do we need this?

src/pgroonga.c
```diff
@@ -830,6 +831,12 @@ PGrnEnsureLatestDB(void)
 		return false;
 	}
 
+	if (PGrnNScanOpaques != 0)
+	{
+		PGRN_TRACE_LOG_EXIT();
+		return false;
+	}
+
```
kou: Should we do this in `PGrnEnsureLatestDB()`?

Should we do this in `pgroonga_insert()` instead?

src/pgroonga.c
```diff
@@ -5184,7 +5184,10 @@ pgroonga_insert(Relation index,
 						tag)));
 	}
 
-	PGrnEnsureLatestDB();
+	if (PGrnNScanOpaques == 0)
```
kou, line 5212: We should add a comment why we need this condition like we did in other callbacks.

komainu8: I see.
I added it.

PR #986, opened 2026-06-30
test yum: pin PGroonga version before the upgrade test

komainu8:
In the maintenance branch for 3.2.2, both 3.2.2-1 and 3.2.2-2 exist.
We should verify that upgrading from 3.2.2-1 to 3.2.2-2 works correctly.
Therefore, pin PGroonga version before running upgrade test.

### DIFF
--- a/packages/yum/test.sh
+++ b/packages/yum/test.sh
@@ -183,6 +183,10 @@ ${DNF} remove -y ${pgroonga_package}
 #     exit
 #     ;;
 # esac
+
+# In the maintenance branch for 3.2.2, both 3.2.2-1 and 3.2.2-2 exist.
+# We should verify that upgrading from 3.2.2-1 to 3.2.2-2 works correctly.
+# Therefore, pin PGroonga version before running upgrade test.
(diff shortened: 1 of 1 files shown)


packages/yum/test.sh
```diff
@@ -184,7 +184,11 @@ ${DNF} remove -y ${pgroonga_package}
 #     ;;
 # esac
 
-${DNF} install -y ${pgroonga_package}
+# In the maintenance branch for 3.2.2, both 3.2.2-1 and 3.2.2-2 exist.
+# We should verify that upgrading from 3.2.2-1 to 3.2.2-2 works correctly.
+# Therefore, pin PGroonga version before running upgrade test.
+pgroonga_package_suffix=3.2.2-1.el8.x86_64
```
abetomo: `pgroonga_old_release` みたいな名前のほうが意味があって良さそうです。

komainu8: Thanks!
I've fixed it.

abetomo (approved):

PR #993, opened 2026-07-24
ci windows: use old CMake temporary

komainu8:
This is a workaround for the following error:

```
  Run-time dependency groonga found: NO  (tried cmake)
  ERR:
  The --no-warn-unused-cli option is deprecated.  Use -Wno-unused-cli instea
  CMake Error: This should not have happened. If you see this message, you a
  CMake Error at C:/Program Files/CMake/share/cmake-4.4/Modules/Internal/Che
    Failed to configure test project build system.
  Call Stack (most recent call first):
    C:/Program Files/CMake/share/cmake-4.4/Modules/CheckCSourceCompiles.cmak
    C:/Program Files/CMake/share/cmake-4.4/Modules/FindThreads.cmake:160 (ch
    C:/Program Files/CMake/share/cmake-4.4/Modules/FindThreads.cmake:226 (_t
    C:/Program Files/CMake/share/cmake-4.4/Modules/CMakeFindDependencyMacro.
    C:/Program Files/CMake/share/cmake-4.4/Modules/CMakeFindDependencyMacro.
    D:/a/pgroonga/pgroonga-4.0.7-postgresql-14-x64/lib/cmake/Groonga/Groonga
    CMakeLists.txt:20 (find_package)
```

We can remove this when the Meson v1.11.3 is released. This error occurs due to a Meson issue.
See: https://gitlab.kitware.com/cmake/cmake/-/work_items/27971

This problem only occurs with CMake 4.4.0. So, We use CMake 4.3.4 temporary.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -60,6 +60,31 @@ jobs:
           matrix.visual-studio-package
         run: |
           choco install -y ${{ matrix.visual-studio-package }}
+      # This is a workaround for the following error:
+      #
+      #   Run-time dependency groonga found: NO  (tried cmake)
+      #   ERR:
+      #   The --no-warn-unused-cli option is deprecated.  Use -Wno-unused-cli instead.
+      #   CMake Error: This should not have happened. If you see this message, you are probably using a broken CMakeLists.txt file or a problematic release of CMake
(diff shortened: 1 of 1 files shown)


komainu8:
Do you have any comments or suggestions?

abetomo (approved):

